PR #6, opened 2014-04-03
Add view with reference number self-checks.

phgross:
@lukasgraf


jone:
what's the state here? it's almost a year old :wink:

deiferni:
after almost two years this should be ripe now :wink:

PR #12, opened 2015-01-26
Add view to check for reference number prefix collisions.

lukasgraf:
This view checks all repository folders (in all repository roots) for colliding reference number 

(Collision: Two repository folders on the same nesting level having the same reference number prefix)

@deiferni @phgross


deiferni:
:+1:

PR #14, opened 2015-03-09
Add view to list principals used in role assignments.

lukasgraf:
This view returns all **unique principals that are used in role assignments** (below the adapted context).

Also includes the Plone site (if called on site root) and any principals from global role assignments in `portal_role_manager`.

@phgross @deiferni

### DIFF
--- a/opengever/maintenance/browser/list_role_principals.py
+++ b/opengever/maintenance/browser/list_role_principals.py
@@ -1,0 +1,61 @@
+from five import grok
+from plone import api
+from Products.CMFPlone.interfaces import IPloneSiteRoot
+from zope.component.hooks import getSite
+from zope.interface import Interface
+
+
+def get_principals_from_local_roles(obj):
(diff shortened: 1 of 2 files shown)


opengever/maintenance/browser/list_role_principals.py
```diff
@@ -0,0 +1,61 @@
...
+    role_manager = site.acl_users.portal_role_manager
+    principal_roles = list(role_manager._principal_roles.items())
+    principals = [assignment[0] for assignment in principal_roles]
+    return principals
+
+
+def get_all_role_principals(context):
+    path = '/'.join(context.getPhysicalPath())
+    catalog = api.portal.get_tool('portal_catalog')
+    brains = catalog(path=path)
```
phgross: use `unrestrictedSearchResults` to include trashed documents ...

lukasgraf:
@phgross nice catch! Updated...

phgross:
:+1:

PR #18, opened 2015-04-14
Use conditional import in @@health-check view to make it work for og.core < 4.2.

lukasgraf:
`create_session` was moved from `opengever.ogds.base.utils` to `opengever.base.model` in `opengever.core >= 4.2`. This conditional import makes sure `opengever.maintenance` works with all deployed versions of `opengever.core`.

@phgross @deiferni


deiferni:
:+1:

PR #22, opened 2015-06-11
Add testing setup to test against supported versions of opengever.core

lukasgraf:
This adds a basic testing setup to test against the currently supported versions of `opengever.core`:

- `2.7`
- `2.8`
- `2.9`
- `3.0`
- `3.1`
- `3.2`
- `3.3`
- `develop`

Also includes a simple testing layer based on the `COMPONENT_REGISTRY_ISOLATION` layer that loads the ZCML for `opengever.maintenance` to trigger any import errors that would prevent the instance from booting.

Many of these tests will currently **fail** because we haven't yet addressed the problems they're supposed to uncover.

### DIFF
--- a/test-og-develop.cfg
+++ b/test-og-develop.cfg
@@ -1,0 +1,16 @@
+[buildout]
+
+extends =
+    https://raw.githubusercontent.com/4teamwork/opengever.core/master/sources.cfg
+    https://raw.github.com/4teamwork/ftw-buildouts/master/test-plone-4.3.x.cfg
+    http://kgs.4teamwork.ch/release/opengever/develop
+    base-testing.cfg
+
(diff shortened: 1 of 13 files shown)


test-og-develop.cfg
```diff
@@ -0,0 +1,16 @@
+[buildout]
+
+extends =
+    https://raw.githubusercontent.com/4teamwork/opengever.core/master/sources.cfg
```
lukasgraf: This will eventually have to be modified to use `buildout-proxy` once we make the `opengever.core` repo private again.

deiferni:
aus meiner Sicht :+1: 

Ziel erreicht, die tests failen :wink:

PR #23, opened 2015-06-11
Restore compatibility with all supported og.core versions

lukasgraf:
This PR introduces a `compat` module that should be used for all conditional imports, compatibility hacks etc... to ensure the `opengever.maintenance` package is compatible with all supported versions of `opengever.core`.

"Compatible" in this context means that it doesn't produce errors (like `ImportErrors` or other ZCML/Grok related errors) **on startup** that would result in the instance not booting. Whether or not it's worth to make the actual *functionality* compatible with all versions will have to be determined on a case-by-case basis.

@deiferni @phgross

### DIFF
--- a/opengever/maintenance/compat.py
+++ b/opengever/maintenance/compat.py
@@ -1,0 +1,25 @@
+"""
+Compatibility module to help support all relevant versions of opengever.core
+"""
+
+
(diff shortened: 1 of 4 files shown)


deiferni:
:+1:

PR #24, opened 2015-07-27
Fix bin/instance run for recent plone.recipe.zope2instance versions

lukasgraf:
Add a fake `-c` option to work around an issue with recent versions of `plone.recipe.zope2instance`'s `bin/interpreter` script.

The `bin/interpreter` script fails at argument parsing and passes a `-c` to the argument parser of `bin/instance run` scripts.

See https://dev.plone.org/ticket/13414 for details.
@deiferni @phgross

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,11 @@ Changelog
 1.0 (unreleased)
 ----------------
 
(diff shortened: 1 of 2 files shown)


phgross:
Kannst du hier nochmals rebasen? die tests sollten nun passen.

PR #26, opened 2015-08-24
Add script to find all Dossiers missing a reference_number.

lukasgraf:
This script should find all dossiers that are missing a reference number.

@phabegger

### DIFF
--- a/opengever/maintenance/scripts/check_for_missing_refnum.py
+++ b/opengever/maintenance/scripts/check_for_missing_refnum.py
@@ -1,0 +1,57 @@
+from Acquisition import aq_parent
+from opengever.base.interfaces import IReferenceNumberPrefix
+from opengever.dossier.behaviors.dossier import IDossierMarker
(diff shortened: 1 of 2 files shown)


opengever/maintenance/scripts/check_for_missing_refnum.py
```diff
@@ -0,0 +1,57 @@
...
+        ann = IAnnotations(parent)
+
+        if DOSSIER_KEY not in ann:
+            print "Key '{}' not found in annotations for {}".format(
+                DOSSIER_KEY, parent.absolute_url())
+            continue
+
+        mapping = ann[DOSSIER_KEY]
+        if PREFIX_REF_KEY not in mapping:
+            print "Key '{}'' not found in annotations for {}".format(
```
phgross: Nur ein Detail aber `"Key '{}' not` anstatt `"Key '{}'' not`

phgross:
Sieht gut aus :+1: .

PR #27, opened 2015-08-27
Add script to update the PW for a user in the ZODBUserManager.

lukasgraf:
Only handles users in the `ZODBUserManager` on the Zope Application Root.

@phgross @deiferni


phgross:
:+1:

PR #31, opened 2015-09-10
Add script to find tasks with title > 255 chars.

lukasgraf:
This script will identify Plone tasks whose title is longer than 255 characters and output a list of them in CSV format.

@deiferni


deiferni:
:+1:

PR #34, opened 2015-09-15
Add script to discover inconsistencies between the Plone Lexicon and ZCTextIndexes

lukasgraf:
This script attempts to discover inconsistencies between the Plone Lexicon and ZCTextIndexes.

It should detect cases where a ZCTextIndex references a word (by an integer word ID (`wid`)) that doesn't exist in the lexicon (any more). It does this by trying to fetch all indexed data, for all indexes, for all brains.

In the case of ZCTextIndexes, this will cause `getEntryForObject` to look up a list of word IDs in the lexicon, and trigger a KeyError if that word isn't in the lexicon.

It seems that the only reliable way to fix this situation is to first **CLEAR** all the affected indexes, and then rebuild them.

An logfile of the first analysis done with this script is at `0000oglx13:~/debug_logs/rid_key_errors_analysis_2015-09-15.log`

The fix for the corresponding upgrade-step is in 4teamwork/opengever.core#1224

@phgross @deiferni

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,10 @@ Changelog
 1.0 (unreleased)
 ----------------
 
+- Add script to discover inconsistencies between the Plone Lexicon
+  and ZCTextIndexes.
+  [lgraf]
+
 - Add script to crop long dossier and task titles.
   [lgraf]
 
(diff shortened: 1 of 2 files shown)


deiferni:
:+1:

PR #35, opened 2015-09-23
Add scripts to fix mail migrations.

deiferni:
When migrating the class of contenttypes one must ensure that all the references to instances are updated/changed as well, see https://www.fourdigits.nl/blog/changing-your-packagename#1411464714874959. This PR adds two scripts to recover from incorrectly migrated mail classes and make sure that all references are up to date as well.

- fix_mail_class_intids: fixes references from the `intid` utility and should theoretically migrate all broken mails
- fix_mail_class_query: due to historical reasons, it was used first to recover from a failed migration but did not update the underlying problem, i.e. references from the intid utility. It should be run after above script to make sure that all is well.

See https://github.com/4teamwork/opengever.core/issues/1215.

### DIFF
--- a/opengever/maintenance/scripts/fix_mail_class_intids.py
+++ b/opengever/maintenance/scripts/fix_mail_class_intids.py
@@ -1,0 +1,105 @@
+"""
+A bin/instance run script which checks and fixes the class of
+all mail objects that are references from the intid catalog.
+
+bin/instance0 run fix_mail_class_intids.py
+"""
+
+from Acquisition import aq_base
+from Acquisition import aq_parent
+from ftw.mail.mail import Mail
+from opengever.mail.mail import OGMail
+from opengever.maintenance.debughelpers import setup_app
+from opengever.maintenance.debughelpers import setup_plone
+from Products.BTreeFolder2.BTreeFolder2 import BTreeFolder2Base
+from zope.app.intid.interfaces import IIntIds
+from zope.component import getUtility
+from zope.interface import directlyProvidedBy
+from zope.interface import directlyProvides
+import transaction
+
+
+def find_key_references_with_wrong_class(intids):
+    wrong_ids = [keyref for keyref in intids.ids
+                 if keyref.object.__class__ == Mail]
+    wrong_refs = [keyref for keyref in intids.refs.values()
+                  if keyref.object.__class__ == Mail]
+    assert set(wrong_ids) == set(wrong_refs)
+
+    return wrong_refs
+
+
+def fix_intids(intids, mail):
+    """Fix reference in intids btree and keyreference objects.
+    """
+
+    intid = intids.getId(mail)
+    reference_to_persistent = intids.refs[intid]
+
+    del intids.refs[intid]
+    del intids.ids[reference_to_persistent]
+
+    intids.refs[intid] = reference_to_persistent
+    intids.ids[reference_to_persistent] = intid
+
+    reference_to_persistent._p_changed = True
+    return intid
+
+
+def fix_mail_class(plone):
+    catalog = plone.portal_catalog
+    intids = getUtility(IIntIds)
+
+    wrong_refs = find_key_references_with_wrong_class(intids)
+    print '{} wrong references in intid utility'.format(len(wrong_refs))
+
+    for reference in wrong_refs:
+        unwrapped_mail = reference.object
+
+        # query mail from catalog for correct acquisition wrapping
+        brains = catalog.unrestrictedSearchResults(UID=unwrapped_mail.UID())
+        if len(brains) < 1:
+            # some mails seem to be stuck in the intid catalog but cannot be
+            # accessed otherwise. Also migrate them to be on the safe side.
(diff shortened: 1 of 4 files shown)


opengever/maintenance/scripts/fix_mail_class_intids.py
```diff
@@ -0,0 +1,105 @@
...
+all mail objects that are references from from the intid catalog.
+
+bin/instance0 run fix_mail_class_intids.py
+"""
+
+from Acquisition import aq_base
+from Acquisition import aq_parent
+from ftw.mail.mail import Mail
+from opengever.mail.mail import OGMail
+from opengever.maintenance.debughelpers import setup_debug_mode
```
lukasgraf: `setup_debug_mode` shouldn't be used in scripts like this. It's intended for exactly what it says, a helper for setting up the `bin/instance debug` environment. Please use [`setup_app()`](https://github.com/4teamwork/opengever.maintenance/blob/master/opengever/maintenance/debughelpers.py#L56) and [`setup_plone_site()`](https://github.com/4teamwork/opengever.maintenance/blob/master/opengever/maintenance/debughelpers.py#L31) as needed.

deiferni: oh, good to know, much better than grabbing plone from `globals`.

opengever/maintenance/scripts/fix_mail_class_intids.py
```diff
@@ -0,0 +1,105 @@
...
+    intids = getUtility(IIntIds)
+
+    wrong_refs = find_key_references_with_wrong_class(intids)
+    print '{} wrong references in intid utility'.format(len(wrong_refs))
+
+    for reference in wrong_refs:
+        unwrapped_mail = reference.object
+
+        # query mail from catalog for correct acquisition wrapping
+        brains = catalog(UID=unwrapped_mail.UID())
```
lukasgraf: `catalog.unrestrictedSearchResults(`

opengever/maintenance/scripts/fix_mail_class_intids.py
```diff
@@ -0,0 +1,105 @@
...
+        if len(brains) < 1:
+            # some mails seem to be stuck in the intid catalog but cannot be
+            # accessed otherwise. Also migrate them to be on the safe side.
+            unwrapped_mail.__class__ = OGMail
+            unwrapped_mail._ofs_migrated = True
+            unwrapped_mail._p_changed = True
+            fix_intids(intids, unwrapped_mail)
```
lukasgraf: So there are in fact mails that aren't indexed in the catalog? Would it maybe make sense to reindex them here? Or are those maybe trashed mails that have been missed by the above query?

deiferni: I think i should have used `unrestrictedSearchResults`.

There is another script to check for uncatalogued mails which did not find any of them.

opengever/maintenance/scripts/fix_mail_class_query.py
```diff
@@ -0,0 +1,84 @@
...
+
+    counter = 0
+    mails_to_fix = []
+
+    mails = catalog.unrestrictedSearchResults(
+        {"portal_type": 'ftw.mail.mail'})
+
+    for brain in mails:
+        mail = brain.getObject()
+        if mail.__class__ == Mail:
```
lukasgraf: Hmm. Thinking about `__class__ == Mail` vs. `__class__ != OGMail`. In fact, I think I'd even prefer `__class__.__name__ != 'OGMail'` - then we could absolutely rule out the check failing because of different versions of the persisted class.

deiferni: Hmm it could only fail when there are multiple instances of one class are floating around, right?. Well you never know, i'll adapt to string comparison as suggested to be on the safe side :+1:.

lukasgraf: Well, it depends on how exactly the mechanism for unpickling persistent classes works. To my knowledge, the simplified version is `Turn the class into a dotted name, pickle the dotted name as a string, resolve the dotted name to c class after unpickling`. But I haven't looked at the actual code, so I was thinking better safe than sorry ;)

lukasgraf:
Other than the commented points: Looks good to me! :+1: Nice catch BTW regarding those intid references :cake:  

Here's hoping we can finally lay this issue to rest after this... :checkered_flag:

deiferni:
@lukasgraf @phgross updated

deiferni:
rebased.

opengever/maintenance/scripts/fix_mail_class_intids.py
```diff
@@ -0,0 +1,105 @@
+"""
+A bin/instance run script which checks and fixes the class of
+all mail objects that are references from from the intid catalog.
```
phgross: from from

deiferni:
@lukasgraf @phgross updated

PR #37, opened 2015-10-01
Add browser view to register logging filters for the SQLAlchemy logger

lukasgraf:
This view allows to register a filter for the SQLAlchemy logger that greatly reduces it's verbosity if SQLAlchemy logging is enabled with `echo=True`.

For now it registers a filter that only logs:
- `UPDATE tasks` statements
- The parameters immediately following that statement 

Usage:
- Enable SQLAlchemy logging on engine level, for example:

``` xml
    <db:engine name="opengever.db" echo="true"
      url="..." pool_recycle="3600"/>
    <db:session name="opengever" engine="opengever.db"/>
```
- Visit the `@@enable-sqlalchemy-logging-filter` view on the Plone Site root.

@phgross @deiferni

### DIFF
--- a/opengever/maintenance/browser/enable_sqlalchemy_logging_filter.py
+++ b/opengever/maintenance/browser/enable_sqlalchemy_logging_filter.py
@@ -1,0 +1,58 @@
+from five import grok
+from opengever.ogds.base.utils import create_session
+from Products.CMFPlone.interfaces import IPloneSiteRoot
+from threading import RLock
+import logging
+
+
+# Module globals to track whether the next bound SQL parameters should be
+# logged or not, and the according lock to ensure thread safety
+_log_next_param = False
+_log_next_param_lock = RLock()
+
+
+class TaskUpdateFilter(logging.Filter):
+    """Filters records so that only UPDATE statements for the 'task' table
+    and the corresponding parameters are logged
+    """
+
+    @staticmethod
+    def filter(record):
+        global _log_next_param
+
+        if _log_next_param and TaskUpdateFilter.is_param_entry(record):
+            with _log_next_param_lock:
+                _log_next_param = False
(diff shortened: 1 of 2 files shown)


opengever/maintenance/browser/enable_sqlalchemy_logging_filter.py
```diff
@@ -0,0 +1,58 @@
...
+from opengever.ogds.base.utils import create_session
+from Products.CMFPlone.interfaces import IPloneSiteRoot
+from threading import RLock
+import logging
+
+
+# Module globals to track whether the next bound SQL parameters should be
+# logged or not, and the according lock to ensure thread safety
+_log_next_param = False
+_log_next_param_lock = RLock()
```
deiferni: I don't understand why we need to share this information globally, i.e. between threads? If not i think it would be better to use threading.local? e.g.:

```
logging_state = threading.local()

...
class TaskUpdateFilter(logging.Filter):
    def __init__(self):
        logging_state.log_next_param = False

    def filter(record):
        if logging_state.log_next_param and TaskUpdateFilter.is_param_entry(record):
            logging_state.log_next_param = False
            return True
    ...

```

*untested code, but something like that should work :wink:*

deiferni:
@lukasgraf i did a scan for open PRs and stumbled upon this, is it still relevant?

opengever/maintenance/browser/enable_sqlalchemy_logging_filter.py
deiferni: As discussed, this seems rather theoretical and should not really matter in practice. We'll leave the code as is for the moment :wink:.

lukasgraf:
Not highly relevant anymore, will reopen if needed again.

PR #41, opened 2015-11-10
Added maintenance script to rename ids of repositoryfolders,

phgross:
when they no longer correlate with the title.

Used for the AI [maintenance job](https://extranet.4teamwork.ch/support/gever-ai/tracker/113).

@lukasgraf @deiferni

### DIFF
--- a/opengever/maintenance/scripts/rename_repository_folders.py
+++ b/opengever/maintenance/scripts/rename_repository_folders.py
@@ -1,0 +1,105 @@
+"""
+Script to rename the id of repository folders, when the id does no longer
+correlate with the title.
+
+USAGE: bin/instance run <path_to>/rename_repository_folders.py
+OPTIONS:
+ -v verbose
+ -n dry run
+ -s site root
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/rename_repository_folders.py
```diff
@@ -0,0 +1,105 @@
...
+    to_fix = check_ids()
+    for item in to_fix:
+        rename(item.get('obj'), item.get('new_id'))
+        reindex_task_path(item.get('obj'))
+
+        if options.verbose:
+            print 'Renamed: {}'.format(item)
+
+        if not options.dry_run:
+            transaction.commit()
```
lukasgraf: Do you really want to commit after every item? This should be moved out one indentation level I think.

phgross: As discussed. Yes.

lukasgraf:
:+1: - please rebase.

PR #42, opened 2015-11-17
Added Script to fix broken sortable title indexes

phgross:
Find brains which has wrongly not been reindexed by the [plone.app.upgrade.v43.reindex_sortable_title](https://github.com/plone/plone.app.upgrade/blob/master/plone/app/upgrade/v43/alphas.py#L11) and reindex them.

Additionally I've dropped the support for GEVER versions < 3.0

@lukasgraf @deiferni 

Extranet: https://extranet.4teamwork.ch/extern/opengever-kanton-zug/sprint-backlog/20255

### DIFF
--- a/opengever/maintenance/scripts/fix_broken_sortable_title_indexes.py
+++ b/opengever/maintenance/scripts/fix_broken_sortable_title_indexes.py
@@ -1,0 +1,90 @@
+from datetime import datetime
+from opengever.maintenance.debughelpers import setup_app
+from opengever.maintenance.debughelpers import setup_option_parser
+from opengever.maintenance.debughelpers import setup_plone
+from plone import api
+from Products.CMFPlone.CatalogTool import MAX_SORTABLE_TITLE
+import logging
+import re
+import transaction
+
+logger = logging.getLogger('opengever.maintenance')
+SEPARATOR = '-' * 78
+
+# Finds only string containing digits, but was not found by the regex of the
+# plone upgrade (`plone.app.upgrade.v43.alphas.num_sort_regex`)
+num_sort_regex = re.compile('\D+\d+')
+six_digits_regex = re.compile('\d{6}')
+
+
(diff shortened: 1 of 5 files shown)


opengever/maintenance/scripts/fix_broken_sortable_title_indexes.py
```diff
@@ -0,0 +1,90 @@
...
+from opengever.maintenance.debughelpers import setup_app
+from opengever.maintenance.debughelpers import setup_option_parser
+from opengever.maintenance.debughelpers import setup_plone
+from plone import api
+from Products.CMFPlone.CatalogTool import MAX_SORTABLE_TITLE
+import logging
+import re
+import transaction
+
+logger = logging.getLogger('opengever.activity')
```
lukasgraf: Nope.

opengever/maintenance/scripts/fix_broken_sortable_title_indexes.py
```diff
@@ -0,0 +1,90 @@
...
+        return
+
+    from Products.PluginIndexes.FieldIndex import FieldIndex
+    if not isinstance(sort_title_index, FieldIndex.FieldIndex):
+        logger.warn('Fix script cancelled, sort_title_index is not a FieldIndex.')
+        return
+
+    change = []
+    for i, (name, rids) in enumerate(sort_title_index._index.iteritems()):
+        if len(name) > MAX_SORTABLE_TITLE or num_sort_regex.match(name):
```
lukasgraf: This should still be `.search()` instead of `.match()` - `match` only matches at *the beginning of the string*

lukasgraf: As discussed: Works also with `.match()`, because `\D+\d+` matches one or more non-digits followed by one or more digits.

lukasgraf:
:+1:

PR #43, opened 2015-12-14
Fix tests and add tests for new release 3.5.x

deiferni:
Depends on https://github.com/4teamwork/opengever.core/pull/1467


lukasgraf:
Tests are still failing because the `profilehook` fix is in none of the released versions of `opengever.core` :(

lukasgraf:
:tada: :+1:

PR #45, opened 2016-01-05
Don't check refnums for template dossiers, they are different.

deiferni:
Template dosssiers use their title as local part of the reference number. We don't want to validate those.


lukasgraf:
:+1:

PR #52, opened 2016-02-15
Add check to verify catalog _wordinfo consistency.

### LINKED ISSUE #51: Extend script to check for catalog inconsistencies with check for broken _wordinfo
The [`check_for_rid_key_errors.py`](https://github.com/4teamwork/opengever.maintenance/blob/master/opengever/maintenance/scripts/check_for_rid_key_errors.py) should be extended with a check that verifies that the `index._wordinfo` mapping is consistent.

xref: https://extranet.4teamwork.ch/support/ph-vs/tracker/407

lukasgraf:
During a massive rename operation, we saw `KeyErrors` during unIndexing of objects from the attempt to delete `doc2score[rid]`.

This check tries to force that kind of `KeyError`, by accessing `idx.index._wordinfo[wid][rid]` for each indexed word in each index with a lexicon (usually `ZCTextIndexes`).

The error seemed to be of a transient nature, so it's unclear whether this is actually a persistent problem - but if it is, this check should uncover it.

Fixes #51.

The inconsistency that this check detects can be created with something like this:

```python
from five import grok
from plone import api
from zope.interface import Interface


PATH = 'fd/ordnungssystem/fuehrung/gemeinderecht/dossier-15/document-24'
IDX_NAME = 'SearchableText'


class BreakWordinfoView(grok.View):

    grok.name('break-wordinfo')
    grok.context(Interface)
    grok.require('cmf.ManagePortal')

    def render(self):
        raise Exception("Don't run me in production!")

        portal = self.context
        catalog = api.portal.get_tool('portal_catalog')
        doc = portal.restrictedTraverse(PATH)
        brain = catalog(path='/'.join(doc.getPhysicalPath()))[0]

        # Get index and AQ wrap it
        idx = catalog._catalog.indexes[IDX_NAME]
        idx = idx.__of__(catalog)
        lexicon = idx.getLexicon()

        rid = brain.getRID()
        index_data = catalog.getIndexDataForRID(rid)

        # Get first index word for SearchableText
        word = index_data[IDX_NAME][0]
        wid = lexicon._wids[word]

        doc2score = idx.index._wordinfo[wid]

        # Delete the score for that word in that document
        del doc2score[rid]

        if doc2score:
            idx.index._wordinfo[wid] = doc2score
        idx.index._wordinfo._p_changed = True

        msg = "Removed score for word %r (%r) in doc with RID %r (%s) " \
              "from index %r" % (word, wid, rid, PATH, IDX_NAME)
        print msg
        return msg
```


lukasgraf:
@deiferni please review

deiferni:
:+1:

PR #56, opened 2016-03-08
Commit in chunks when renaming 

deiferni:
This PR improves the script to rename document its by:
- Committing every 100 items
- Adding Progress-Log
- Removing too verbose logging, renamed items are written to a .csv file already, that is sufficient


deiferni:
@lukasgraf plz have a look

lukasgraf:
Looks good to me :+1:

PR #61, opened 2016-08-23
Move bumblebee installation script from og.core to og.maintenace.

phgross:
The `scripts` directory from `opengever.core` is not part of the egg, therefore it makes more sense to add the bumblebee installation script to the opengever.maintenance package.

@deiferni


deiferni:
👌

PR #62, opened 2016-08-24
Add script to move documents in proposals.

deiferni:
This was only allowed very briefly during development. There is no
upgrade step. Some cases seem to have sneaked into our staging environment,
so we fix these by moving the affected documents to the proposals dossier.

Closes https://github.com/4teamwork/opengever.core/issues/2084.


deiferni:
@phgross

PR #76, opened 2017-03-25
Add intermediate commit option to bumblebee activation.

deiferni:
Add option to commit every N elements. This currently has only an effect when run with `-m history`.

### DIFF
--- a/opengever/maintenance/scripts/bumblebee_installation.py
+++ b/opengever/maintenance/scripts/bumblebee_installation.py
@@ -82,6 +82,11 @@ parser.add_option("-p", "--plone", dest="plone_path",
 parser.add_option("-c", "--config", dest="config",
                   help="Zope-Config (do not use this)")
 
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/bumblebee_installation.py
```diff
@@ -139,6 +144,11 @@ def main(app, argv=sys.argv[1:]):
                 annotations[DOCUMENT_CHECKSUM_ANNOTATION_KEY] = version_checksum
                 archived_obj._p_changed = True
 
+            if options.intermediate_commit:
+                if index % options.intermediate_commit == 0:
```
lukasgraf: Just FYI, this will also commit at 0 (zero) items, at the very start of the loop. Usually that's not an issue, but if that's undesired behavior the condition would need to be tweaked.

deiferni: @lukasgraf it should not matter, but it is nicer when we don't 😉 . updated.

PR #77, opened 2017-03-25
Use intermediate commits during reindex if available.

deiferni:
Use intermediate commit of `ftw.bumblebee` during reindex, if available. See https://github.com/4teamwork/ftw.bumblebee/pull/111 for more information.

_This will also work if intermediate_commit is not specified on the command line._

### DIFF
--- a/opengever/maintenance/scripts/bumblebee_installation.py
+++ b/opengever/maintenance/scripts/bumblebee_installation.py
@@ -85,7 +85,8 @@ parser.add_option("-c", "--config", dest="config",
 parser.add_option("-i", "--intermediate-commit", dest="intermediate_commit",
                   default=None, type="int",
                   help="Intermediate commit every n processed elements. "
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/bumblebee_installation.py
```diff
@@ -111,7 +111,14 @@ def main(app, argv=sys.argv[1:]):
 
     if mode == 'reindex':
         LOG.info("Start indexing objects...")
-        converter.reindex()
+        try:
+            converter.reindex(intermediate_commit=options.intermediate_commit)
+        except TypeError:
+            if options.intermediate_commit:
```
lukasgraf: Should probably remove the `"Only has an effect on '-m history' at the moment."` from the options parser

deiferni: yes indeed, thx! amended.

PR #78, opened 2017-03-31
Bumblebee installation script: Skip documents with more than 50 versions

lukasgraf:
Bumblebee installation script: Skip documents with more than 50 versions when calculating checksums for versions, and log them.

### DIFF
--- a/opengever/maintenance/scripts/bumblebee_installation.py
+++ b/opengever/maintenance/scripts/bumblebee_installation.py
@@ -88,6 +88,12 @@ parser.add_option("-i", "--intermediate-commit", dest="intermediate_commit",
                        "Works with '-m history' and '-m reindex' at the "
                        "moment.")
 
+parser.add_option("--skip-many-versions",
+                  action="store_true",
+                  default=False,
+                  help="Skip objects with many (>50) versions when "
+                       "calculating checksums for versions using -m history")
+
 
 def main(app, argv=sys.argv[1:]):
     options, args = parser.parse_args()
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/bumblebee_installation.py
```diff
@@ -130,12 +130,21 @@ def main(app, argv=sys.argv[1:]):
...
             obj = brain.getObject()
             versions = repository.getHistory(obj)
+
+            if len(versions) > 50:
+                LOG.info(
+                    'Skipping object with more than 50 versions: %s' % obj.id)
+                skipped.append((obj.id, obj.absolute_url()))
+                continue
+
+            LOG.info('  Calculating version checksums for %s' % obj.id)
```
lukasgraf: Indented by two spaces to make it easier to distinguish the per-object logging (which can get quite noisy) from the per-batch logging.

opengever/maintenance/scripts/bumblebee_installation.py
```diff
@@ -157,7 +166,15 @@ def main(app, argv=sys.argv[1:]):
                     LOG.info("Committing at {} documents.".format(index))
                     transaction.commit()
 
-        return transaction.commit()
+        transaction.commit()
+
+        if skipped:
+            LOG.warn('The following objects have been skipped:')
+            for obj_id, obj_url in skipped:
+                LOG.info("%s - %s" % (obj_id, obj_url))
```
lukasgraf: Can't use log level WARN for those because the log handler fuckery at the top of the module causes everything above INFO to be logged twice.

lukasgraf:
@deiferni updated, tested again for both code paths.

PR #79, opened 2017-05-16
Add script to export ordnungssystem into xlsx file

raphael-s:
The scripts exports the data from the ordnungssystem into an xlsx file. After running the script, the data is availabe in the  `ordnungssystem.xlsx` file.
The script has to be executed like this: `bin/instance -O fd run scripts/get_ordnungssystem_xlsx.py`


phgross:
No longer necessary (since https://github.com/4teamwork/opengever.core/pull/4408)

PR #81, opened 2017-05-30
Canonicalize the dev version in order to squelch a setuptools warning

Rotonen:
Ref. [PEP-440](https://www.python.org/dev/peps/pep-0440/#implicit-development-release-number).


lukasgraf:
*sigh* (I'm looking at you, setuptools) - good call

PR #83, opened 2017-05-31
There are no .txt files in the project root to package into the egg

Rotonen:
There are no .txt files in the project root.


Rotonen:
@deiferni the point is to reduce buildout noise. This takes out unnecessary `setup.py develop` output.

PR #87, opened 2017-07-31
Remove includeDependencies directive.

jone:
The includeDependencies directive is used for including the ZCML of dependencies, which do not provide a z3c.autoinclude entry point. Nowadays most packages provide the entry point.

The includeDependencies directive is including all dependencies. This is not explicit and it may load a lot of ZCML.

This removes the includeDependencies directive and loads five.grok explicitly.


jone:
Maybe its not a good idea.

PR #89, opened 2017-08-17
Add script that checks for users that are present in OGDS and LDAP with IDs that differ ONLY by case

lukasgraf:
Add script that checks for users that are present in OGDS and LDAP with IDs that differ ONLY by case and lists them in a mapping.

(This is used to produce the mapping used for the AI usermigration).

### DIFF
--- a/opengever/maintenance/scripts/list_userids_differing_in_case.py
+++ b/opengever/maintenance/scripts/list_userids_differing_in_case.py
@@ -1,0 +1,74 @@
+"""
+This script checks for users that are present in OGDS and LDAP with IDs
+that differ ONLY by case and lists them in a mapping.
+
+(This is used to produce the mapping for migrating users that changed casing,
+e.g. mixed case -> lowercase user IDs.)
+"""
+
+from opengever.base.model import create_session
+from opengever.maintenance.debughelpers import setup_app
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/list_userids_differing_in_case.py
```diff
@@ -0,0 +1,73 @@
+"""
+This script checks for users that are present in OGDS and LDAP with IDs
+that differ ONLY by case and lists them in a mapping.
+
+(This is used to produce the mapping used for the AI usermigration).
```
deiferni: i would not mention AI here, this could be used in any user-migration?

opengever/maintenance/scripts/list_userids_differing_in_case.py
```diff
@@ -0,0 +1,73 @@
...
+
+def case_insensitive_get(seq, key):
+    matches = [item for item in seq if item.lower() == key.lower()]
+    assert len(matches) in (0, 1)
+    if matches:
+        return matches[0]
+
+
+def generate_usermigration_mapping(plone):
+    """Produce mapping for the AI user migration (lowercased users).
```
deiferni: same here

PR #90, opened 2017-09-12
Add script to fix broken journal entries after OGGBundle import.

phgross:
Fixes encoding for zope.i18nmessageid.Message mapping values, for journal entries of objects which has been created with an OGGBundle import.

See https://github.com/4teamwork/opengever.core/pull/3369 for more details.

### DIFF
--- a/opengever/maintenance/scripts/fix_journal_message_mappings.py
+++ b/opengever/maintenance/scripts/fix_journal_message_mappings.py
@@ -1,0 +1,102 @@
+from ftw.journal.config import JOURNAL_ENTRIES_ANNOTATIONS_KEY
+from ftw.journal.interfaces import IAnnotationsJournalizable
+from ftw.upgrade import ProgressLogger
+from opengever.maintenance.debughelpers import setup_app
+from opengever.maintenance.debughelpers import setup_option_parser
+from opengever.maintenance.debughelpers import setup_plone
+from plone import api
+from zope.annotation.interfaces import IAnnotations
+from zope.i18nmessageid import Message
+import logging
+import sys
+import transaction
+
+
+logger = logging.getLogger('opengever.maintenance')
+handler = logging.StreamHandler(stream=sys.stdout)
+logging.root.addHandler(handler)
+logging.root.setLevel(logging.INFO)
+
+MAY_NEED_FIXING = ['title', 'repository']
+
+
+class JournalMessageFixer(object):
+    """Fixes encoding for zope.i18nmessageid.Message mapping values, for
+    journal entries of objects which have been created with an OGGBundle
+    import.
+
+    See https://github.com/4teamwork/opengever.core/pull/3369 for more details.
+    """
+
+    def __init__(self, portal, options):
+        self.portal = portal
+        self.options = options
+        self.counter = 0
+
+    def get_journal_entries(self, obj):
+        """Returns all journal entries for the given object.
+        """
+        if IAnnotationsJournalizable.providedBy(obj):
+            annotations = IAnnotations(obj)
+            return annotations.get(JOURNAL_ENTRIES_ANNOTATIONS_KEY, [])
+
+        return []
+
+    def fix_entries(self, obj):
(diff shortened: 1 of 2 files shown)


opengever/maintenance/scripts/fix_journal_message_mappings.py
```diff
@@ -0,0 +1,98 @@
...
+handler = logging.StreamHandler(stream=sys.stdout)
+logging.root.addHandler(handler)
+logging.root.setLevel(logging.INFO)
+
+MAPPING_KEYS = ['title', 'repository']
+
+
+class JournalMessageFixer(object):
+    """Fixes encoding for zope.i18nmessageid.Message mapping values, for
+    journal entries of objects which has been created with an OGGBundle import.
```
deiferni: has -> have

opengever/maintenance/scripts/fix_journal_message_mappings.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def main():
+    app = setup_app()
+
+    parser = setup_option_parser()
+    (options, args) = parser.parse_args()
+
+    plone = setup_plone(app, options)
+    cleaner = JournalMessageFixer(plone, options)
```
deiferni: cleaner -> fixer?

opengever/maintenance/scripts/fix_journal_message_mappings.py
```diff
@@ -0,0 +1,98 @@
...
+        """
+        if IAnnotationsJournalizable.providedBy(obj):
+            annotations = IAnnotations(obj)
+            return annotations.get(JOURNAL_ENTRIES_ANNOTATIONS_KEY, [])
+
+        return []
+
+    def fix_entries(self, obj):
+        for entry in self.get_journal_entries(obj):
+            message = entry['action'].get('title')
```
deiferni: do we always have an `action`? if no you may wanna do 

```python
message = entry.get('action', {}).get('title')
```
instead.

opengever/maintenance/scripts/fix_journal_message_mappings.py
```diff
@@ -0,0 +1,98 @@
...
+            if message and isinstance(message, Message) and message.mapping:
+                if self.check_message(message):
```
deiferni: check_message -> needs_fixing

opengever/maintenance/scripts/fix_journal_message_mappings.py
```diff
@@ -0,0 +1,98 @@
...
+                    self.fix_message(entry, message)
+
+    def check_message(self, message):
+        for key in MAPPING_KEYS:
```
deiferni: MAPPING_KEYS -> MAY_NEED_FIXING ?

opengever/maintenance/scripts/fix_journal_message_mappings.py
```diff
@@ -0,0 +1,98 @@
...
+        with mapping values in the correct encoding (unicode).
+        """
+        mapping = old_message.mapping
+        for key in mapping.keys():
+            if key in MAPPING_KEYS:
+                value = mapping[key]
+                if not isinstance(value, unicode):
+                    mapping[key] = value.decode('utf-8')
+
+        entry['action']['title'] = Message(
```
deiferni: i don't understand (from the code) why this is necessary, wouldn't it be enough to just replace the mapping values? could you add a short comment?

opengever/maintenance/scripts/fix_journal_message_mappings.py
phgross: Yes, we always have an `action. If not the script should fail, so that I can have a closer look to those objects.

opengever/maintenance/scripts/fix_journal_message_mappings.py
phgross: The docstring of the method describes what it does. I now also add a line why it's necessary.

phgross:
@deiferni updated ...

PR #98, opened 2017-11-16
 Warmup view: Add mode to warm up entire catalog

lukasgraf:
Adds `?mode=catalog` to the warmup view.

This will load all of the **catalog's internal data structures**.

The query string parameter `zctext_indexes=0` can be appended to avoid loading `ZCTextIndexes`, or `lexicons=0` to just avoid loading their lexicons.

Some **cache related statistics** will be displayed before and after the catalog warmup (total estimated cache size in bytes, number of objects in cache, current instance RSS).

In addition, the view will now run with **elevated privileges** (in order to be able to load actual objects, which are restricted by security, unlike catalog results), and `unrestrictedSearchResults` is being used for catalog queries.

I also dropped the `full` mode which was simply overkill (basically loading every single object on a GEVER instance).

----

Some stats from the `ogg100k` deployment:

`portal_catalog` ZEXP Size: `573M` (250k: 1.6G)

Catalog warmup (with lexicons):

```
2017-11-16T21:45:21 INFO opengever.maintenance Cache size (# objs) (before warmup): 82
2017-11-16T21:45:21 INFO opengever.maintenance Total estimated cache size (before warmup): 0.08 MiB

2017-11-16T21:58:04 INFO opengever.maintenance Cache size (# objs) (after warmup): 2170379
2017-11-16T21:58:04 INFO opengever.maintenance Total estimated cache size (after warmup): 999.25 MiB

2017-11-16T21:58:04 INFO opengever.maintenance Catalog warmup took 762.75s.
```

RSS of instance after warmup: 3576.8 MiB (264.5 before) => **+3312.3 MiB**
Warmup time: `12m 42s`

---

Catalog warmup (without lexicons):

```
2017-11-16T22:03:50 INFO opengever.maintenance Cache size (# objs) (before warmup): 82
2017-11-16T22:03:50 INFO opengever.maintenance Total estimated cache size (before warmup): 0.08 MiB

2017-11-16T22:15:51 INFO opengever.maintenance Cache size (# objs) (after warmup): 2047713
2017-11-16T22:15:51 INFO opengever.maintenance Total estimated cache size (after warmup): 908.05 MiB

2017-11-16T22:15:51 INFO opengever.maintenance Catalog warmup took 720.88s.
```

RSS of instance after warmup: 3267.6 MiB (264.4 before) => **+3003.2 MiB**
Warmup time: `12m 1s`

---

Catalog warmup (without ZCTextIndexes):

```
2017-11-17T14:11:10 INFO opengever.maintenance Cache size (# objs) (before warmup): 82
2017-11-17T14:11:10 INFO opengever.maintenance Total estimated cache size (before warmup): 0.08 MiB

2017-11-17T14:22:44 INFO opengever.maintenance Cache size (# objs) (after warmup): 1949361
2017-11-17T14:22:44 INFO opengever.maintenance Total estimated cache size (after warmup): 559.98 MiB

2017-11-17T14:22:44 INFO opengever.maintenance Catalog warmup took 694.73s.
```

RSS of instance after warmup: 2079.65 MiB (264.4 before) => **+1815.25 MiB**
Warmup time: `11m 35s`

### DIFF
--- a/opengever/maintenance/browser/warmup.py
+++ b/opengever/maintenance/browser/warmup.py
@@ -1,4 +1,7 @@
 from Acquisition import aq_base
+from App.config import getConfiguration
+from collections import OrderedDict
+from datetime import datetime
 from functools import wraps
 from opengever.contact.contact import IContact
 from opengever.document.behaviors import IBaseDocument
@@ -22,7 +25,9 @@ from Products.PluginIndexes.interfaces import IPluggableIndex
 from Products.PluginIndexes.KeywordIndex.KeywordIndex import KeywordIndex
 from Products.PluginIndexes.UUIDIndex.UUIDIndex import UUIDIndex
 from Products.ZCTextIndex.ZCTextIndex import ZCTextIndex
+import json
 import logging
+import os
 import threading
 import time
 import transaction
@@ -43,10 +48,11 @@ COMMON_TYPES = [
 
 class CacheStats(object):
 
-    def __init__(self, conn):
+    def __init__(self, conn, settings):
         self.conn = conn
         self.db = conn.db()
         self.stats_by_idx = {}
+        self.settings = settings
 
     def get_rss(self):
         """Returns this process' current RSS (in kb).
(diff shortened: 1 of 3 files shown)


opengever/maintenance/browser/warmup.py
```diff
@@ -30,54 +44,170 @@ class WarmupView(BrowserView):
...
+    """Load internal index data structures by iterating over contents of
+    *BTrees and *TreeSets, causing their respective buckets to be loaded.
+    """
+    if isinstance(index, (FieldIndex, DateIndex, KeywordIndex)):
+        # value -> set of docids
+        for key, value in index._index.items():
+            list(value)
+
+        # docid -> field value
+        list(index._unindex.items())
```
buchi: One could skip warming up the reverse index (`_unindex`) as it shouldn't be used for querying the catalog. But indexing an object requires it. I'm not sure about the performance impact of a cold reverse index when indexing an object.

lukasgraf:
I've updated the view to include detailed per-index statistics, and ran a few tests again.

See this [internal Google spreadsheet (4 sheets, totals and 3 tested deployments)](https://docs.google.com/spreadsheets/d/1HNy6CAG2gdVDXrPBL8JqH9C57Iss_Mk081fTqS6kX60/edit#gid=547624262) for details.

It seems that omitting UnIndexes unfortunately doesn't make much of a dent in the number of objects in the cache (around 10% less). There's only a handful of indexes though that contribute the most to the obj count, by a long shot. If we can limit these to only warmup ones that actually benefit us, that could significantly reduce obj count in the cache.

Thoughts on where to go from here:
- We could make the warmup **adaptive** to the cache size. (Stop warming up once the ring buffer has reached target size).
- Finding out which indexes are useful to warmup is tricky. Maybe the catalog query plan from a long running production instance could give us some insight, or at least pointers as to which indexes we should look at (`catalog.getCatalogPlan()`, `VALUE_INDEXES`?).

lukasgraf:
Superseded by `ftw.monitor`:
https://github.com/4teamwork/ftw.monitor

PR #99, opened 2017-11-24
Catch transient buildout failures easier on local runs

Rotonen:
Buildout still trips up on itself quite often when having to downgrade itself. Catching this early in a non-interactive local test run is IMO useful.


njohner:
@lukasgraf @deiferni I can't tell here, merge or close?

njohner:
I'll close that

PR #100, opened 2017-11-27
Add fix_one_note_mimetype script.

phgross:
With https://github.com/4teamwork/opengever.core/pull/3660 the mimetype for onenote document has been fixed. But the fix in #3660 solves the problem only for new but not for existing documents. Because we don't have an index for the contentType or the filename an upgradestep would be way to longrunning comparing how many documents are effected. Therefore we add this simple opengever.maintenance script. We can run offside the release.

### DIFF
--- a/opengever/maintenance/scripts/fix_one_note_mimetypes.py
+++ b/opengever/maintenance/scripts/fix_one_note_mimetypes.py
@@ -1,0 +1,51 @@
+from opengever.document.document import IDocumentSchema
+from opengever.maintenance.debughelpers import setup_app
+from opengever.maintenance.debughelpers import setup_option_parser
+from opengever.maintenance.debughelpers import setup_plone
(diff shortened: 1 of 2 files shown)


opengever/maintenance/scripts/fix_one_note_mimetypes.py
```diff
@@ -0,0 +1,51 @@
...
+            obj = brain.getObject()
+            if not obj.file:
+                continue
+
+            filename, ext = splitext(obj.file.filename)
+            if ext == '.one':
+                obj.file.contentType = 'application/onenote'
+                obj.reindexObject()
+
+                print 'File {} with {} fixed.'.format(obj, obj.file.filename)
```
deiferni: filename is always ascii only, right? otherwise this may 💥 .

PR #101, opened 2017-11-27
Add a script to persist stored as paper form default values

Rotonen:
* Fires both the Python GC and the site GC every 100 `brain.getObject()`
* Persists unpersisted default values for `preserved_as_paper` to the current registry default

### DIFF
--- a/opengever/maintenance/scripts/persist_default_value_for_paperform.py
+++ b/opengever/maintenance/scripts/persist_default_value_for_paperform.py
@@ -1,0 +1,97 @@
+from logging import Formatter
+from logging import getLogger
+from logging import INFO
+from logging import StreamHandler
+from opengever.base.default_values import object_has_value_for_field
+from opengever.document.behaviors.metadata import IDocumentMetadata
+from opengever.document.behaviors.metadata import preserved_as_paper_default
+from opengever.document.behaviors import IBaseDocument
+from opengever.maintenance.debughelpers import setup_app
+from opengever.maintenance.debughelpers import setup_option_parser
+from opengever.maintenance.debughelpers import setup_plone
+from plone import api
+from time import time
+import gc
+import transaction
+
+
+def fire_gc():
+    start = time()
+    logger.info('Firing garbage collection')
+    api.portal.get()._p_jar.cacheGC()
+    gc.collect()
+    logger.info('Garbage collection done in %.3fs', time() - start)
+
+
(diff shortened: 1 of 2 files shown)


opengever/maintenance/scripts/persist_default_value_for_paperform.py
```diff
@@ -0,0 +1,106 @@
...
+    logger.info('Firing garbage collection')
+    api.portal.get()._p_jar.cacheGC()
+    gc.collect()
+    logger.info('Garbage collection done in %.3fs', time() - start)
+
+
+def persist_default_value_for_paperform():
+    catalog = api.portal.get_tool('portal_catalog')
+    query_filter = {
+        'object_provides': IDocumentSchema.__identifier__,
```
phgross: Needs to be `IBaseDocument`, mails has the same field.

opengever/maintenance/scripts/persist_default_value_for_paperform.py
```diff
@@ -0,0 +1,106 @@
...
+
+    logger.info('Persisting paperform default values on documents')
+
+    for i, brain in enumerate(results):
+        title = brain.Title.decode('UTF-8')
+        title_length = len(title)
+
+        # Keep justifying if something is longer
+        # Likely reaches the usable steady state quick enough
+        if title_length > longest_title_length:
```
phgross: I don't really get the point why you have to store and shorten the title. But it's ok.

Rotonen: For output cleanliness to quickly see what is going on. It's a pet peeve of mine to try to parse output which jumps alignment-wise.

opengever/maintenance/scripts/persist_default_value_for_paperform.py
Rotonen: Will amend.

opengever/maintenance/scripts/persist_default_value_for_paperform.py
phgross: Ah ok.

A simple `title[:100]` would has been enough, but it's fine 😄 .

Rotonen: Threw this out for now as kept hitting unicode bugs for justification. Not trivially solvable on Python 2.

Rotonen:
@phgross ready for rereview.

PR #102, opened 2017-12-07
Blobify namedfiles, i.e. convert NamedFile to NamedBlobFile

deiferni:
Add a script to remove named files and convert them to named blob files. Occasionally we seem to have some `NamedFile` instances left. Those should be a `NamedBlobFile`, so this script will perform the conversion.

Successfully tested on a customers release-test deployment with >100 broken mails.

### DIFF
--- a/opengever/maintenance/scripts/blobify_namedfiles.py
+++ b/opengever/maintenance/scripts/blobify_namedfiles.py
@@ -1,55 +1,74 @@
 """
-Script to remove named files and convert them to blobs. Occasionally we seem
-to have some NamedFile instances lying around. Those should be a NamedBlobFile,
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/blobify_namedfiles.py
```diff
@@ -0,0 +1,55 @@
+"""
+Script to remove named files and convert them to blobs. Occasionally we seem
+to have some NamedFile instances lying around. Those should be a NamedBlobFile,
+so this script will perform the conversion.
+
+    bin/instance run ./scripts/blobify_namedfiles.py.py
```
lukasgraf: `py.py` </nitpick>

PR #104, opened 2018-02-02
Script to set mimetypes for IWorks documents (numbers, pages, keynote).

njohner:
Maintenance script to set the mimetypes for IWorks documents.
This complements https://github.com/4teamwork/opengever.core/pull/3835 where the new mimetypes were introduced.

### DIFF
--- a/opengever/maintenance/scripts/set_iworks_mimetypes.py
+++ b/opengever/maintenance/scripts/set_iworks_mimetypes.py
@@ -1,0 +1,52 @@
+from ftw.upgrade.progresslogger import ProgressLogger
+import mimetypes
+from opengever.maintenance.debughelpers import setup_app
+from opengever.maintenance.debughelpers import setup_option_parser
+from opengever.maintenance.debughelpers import setup_plone
+from os.path import splitext
+from plone import api
+import transaction
+
+
+def set_iworks_mimetype():
+    """Search all MS OneNote documents and fixes the contentType.
+    """
+
+    catalog = api.portal.get_tool('portal_catalog')
+    brains = catalog.unrestrictedSearchResults(
+        portal_type='opengever.document.document')
+    # A getIcon value of 'application.png' indicates that the mimetype
+    # coulnd't be resolved for this document - only consider these
+    brains = filter(lambda b: b.getIcon == 'application.png', brains)
+    for brain in ProgressLogger('Set IWorks mimetypes', brains):
+        obj = brain.getObject()
+        if not obj.file:
+            continue
+        fname = obj.file.filename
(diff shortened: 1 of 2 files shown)


njohner:
Ok done!

opengever/maintenance/scripts/set_iworks_mimetypes.py
```diff
@@ -0,0 +1,50 @@
...
+    """Search all MS OneNote documents and fixes the contentType.
+    """
+
+    catalog = api.portal.get_tool('portal_catalog')
+    brains = catalog.unrestrictedSearchResults(
+        portal_type='opengever.document.document')
+    # A getIcon value of 'application.png' indicates that the mimetype
+    # coulnd't be resolved for this document - only consider these
+    brains = filter(lambda b: b.getIcon == 'application.png', brains)
+    for brain in brains:
```
deiferni: this is potentially long-running, could you use a `ProgressLogger`, see https://github.com/4teamwork/opengever.maintenance/blob/81f586b0eb61a3908bb7a28e6e2375c0ba9affd2/opengever/maintenance/scripts/bumblebee_installation.py#L140 or an example.

opengever/maintenance/scripts/set_iworks_mimetypes.py
```diff
@@ -0,0 +1,50 @@
...
+        obj = brain.getObject()
+        fname = obj.file.filename
```
deiferni: the file may be `None` so we need an `if` for that here.

opengever/maintenance/scripts/set_iworks_mimetypes.py
```diff
@@ -0,0 +1,50 @@
...
+        ext = splitext(fname)[1]
+        if ext in [".numbers", ".pages", ".key"]:
+            obj.file.contentType = mimetypes.guess_type(fname)[0]
+            obj.reindexObject()
```
deiferni: please only reindex the affected indices. reindexing is expensive, especially `SearchableText` as this involves file content.

If there is no index and you're interested in metadata only just pick a cheap index like `id`.

opengever/maintenance/scripts/set_iworks_mimetypes.py
```diff
@@ -0,0 +1,50 @@
...
+            print u'File {} with {} fixed.'.format(obj, obj.file.filename)
```
deiferni: i'd probably omit printing each file separately here and just be happy with what `ProgressLogger` tells us. (Also it would be nicer to use `logging` in such cases, but here i'd just drop it 😉 )

njohner:
Ok done.

PR #113, opened 2018-02-23
 Add script to unindex duplicates from the catalog

deiferni:
We have observed that due to a currently unknown bug certain objects are registered twice in the catalog. This script can be passed a path and then tries to figure out if that object is broken and if so, tries to remove the duplication.

Tries to be extra careful and verbose. Only handles the observed case of exactly one duplicate.

### DIFF
--- a/opengever/maintenance/scripts/unindex_duplicates_by_path.py
+++ b/opengever/maintenance/scripts/unindex_duplicates_by_path.py
@@ -1,0 +1,121 @@
+from opengever.maintenance.debughelpers import setup_app
+from opengever.maintenance.debughelpers import setup_option_parser
+from opengever.maintenance.debughelpers import setup_plone
+from plone import api
+import sys
+import transaction
+
+
+USAGE = """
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/unindex_duplicates_by_path.py
```diff
@@ -0,0 +1,120 @@
...
+    print "Entries in catalog paths: {}".format(pre_paths_length)
+    print "Catalog btree length {}".format(pre_btree_length)
+    print ""
+
+    invalid_rids = [each for each in all_rids if each != correct_rid]
+    if len(invalid_rids) != 1:
+        # we could loop, of course, but currently we have only observed
+        # one invalid entry. so better have a close look again should
+        # we ever have multiple invalid entries.
+        print "Can only handle one incorrect rid, got {}".format(invalid_rids)
```
lukasgraf: This should `sys.exit(1)`, no?

deiferni: Very `True`, fixed.

PR #115, opened 2018-03-13
Add script to fix dossier resolvability.

deiferni:
Provides functionality to fix the resolvability of a dossier with already resolved subdossiers that are now in an 'invalid' state. Invalid means that some of the documents in a dossier have a document date outside its dossiers start-end range. This most likely happens when we generate a journal pdf while closing a dossier with an already defined end-date.

### DIFF
--- a/opengever/maintenance/scripts/fix_dossier_resolvability.py
+++ b/opengever/maintenance/scripts/fix_dossier_resolvability.py
@@ -1,0 +1,77 @@
+"""
+Provides functionality to fix the resolvability of a dossier with already
+resolved subdossiers that are now in an 'invalid' state.
+Invalid means that some of the documents in a dossier have a document date
+outside its dossiers start-end range. This most likely happens when we generate
+a journal pdf while closing a dossier with an already defined end-date.
+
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/fix_dossier_resolvability.py
```diff
@@ -0,0 +1,77 @@
...
+        print "Could not find dossier at {}".format(path)
+        sys.exit(1)
+
+    catalog = api.portal.get_tool('portal_catalog')
+    query = {
+        'path': '/'.join(dossier.getPhysicalPath()),
+        'portal_type': 'opengever.document.document'
+    }
+
+    for brain in catalog(**query):
```
lukasgraf: This should probably be `catalog.unrestrictedSearchResults(**query)` - we might potentially have to deal with trashed documents too.

deiferni: AFAIK the validation rule queries restricted also. But you're right i can't hurt, i'll amend.

PR #117, opened 2018-03-23
Add favorites statistics script.

jone:
In order to find out how many users are actually using favorites, this script dumps simple favorites usage statistics. With the -d parameter the statistics can be dumped to a directory as a json file.

The script prints something like:
```python
{
    "favorites_enabled": true,
    "stats": [
        {
            "max": 5,
            "median": 3.5,
            "min": 2,
            "root_id": "ordnungssystem",
            "total_users": 7,
            "with_favorites": 2
        }
    ]
}
```
and dumps into files like `http-localhost-8080-fd.json` (based on public URL).

The idea is that I can run this script for e.g. all SaaS deployments and dump the results to one directory.

### DIFF
--- a/opengever/maintenance/scripts/statistics_favorites.py
+++ b/opengever/maintenance/scripts/statistics_favorites.py
@@ -1,0 +1,75 @@
+"""Favorites usage statistics.
+
+The goal of this script is to collect usage statistics of repository folder favorites.
+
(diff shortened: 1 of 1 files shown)


jone:
Updated: added `min`, `max`, `median` and removed `mean`.

PR #118, opened 2018-03-23
Add script to fix permissions on docs moved out of private area

lukasgraf:
This script updates object permissions for documents (and mails) that **have been moved out of a user's private area** before the fix from 4teamwork/opengever.core#4092 was in place.

It recognizes documents that have incorrect permissions by the fact that they still have `MemberAreaAdministrator` in their `allowedRolesAndUsers` index but are in the outside of a private area (in the public repository).

### DIFF
--- a/opengever/maintenance/scripts/fix_permissions_on_docs_moved_out_of_private_area.py
+++ b/opengever/maintenance/scripts/fix_permissions_on_docs_moved_out_of_private_area.py
@@ -1,0 +1,68 @@
+"""
+Fixes permissions on documents that have previously been moved out of a
+user's private area into the public repository. See opengever.core#4092
+
+    bin/instance run fix_permissions_on_docs_moved_out_of_private_area.py
+
+"""
+from ftw.upgrade.helpers import update_security_for
+from opengever.maintenance.debughelpers import setup_app
+from opengever.maintenance.debughelpers import setup_option_parser
+from opengever.maintenance.debughelpers import setup_plone
+from plone import api
+from Products.CMFCore.CMFCatalogAware import CatalogAware
+import transaction
+
+
+def fix_permissions_for_moved_documents(options):
+    catalog = api.portal.get_tool('portal_catalog')
+
+    private_root_paths = [b.getPath() for b in
+                          catalog.unrestrictedSearchResults(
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/fix_permissions_on_docs_moved_out_of_private_area.py
```diff
@@ -0,0 +1,68 @@
...
+    for brain in brains:
+        if any([brain.getPath().startswith(p) for p in private_root_paths]):
+            # Skip documents in private areas, we're interested in ones
+            # that have been moved *out* of private areas
+            continue
+
+        print "Affected document: %s" % brain.getPath()
+
+        if not options.dryrun:
+            obj = brain.getObject()
```
jone: The query is executed unrestricted. This means it can theoretically result in brains for which the current user has no view permission. Since `getObject` does a restricted lookup, it will throw an `Unauthorized` exception here.

It is therefore good habbit to combine restricted `searchResults` with `getObject` and `unrestrictedSearchResults` with `restrictedTraverse`.
I'd therefore use:
```python
obj = site.restrictedTraverse(brain.getPath())
```

I'm aware that we run the scripts with manager role and that the manager, in theory, should have access to all the things. But I think it is good habbit to not respect security in such maintenance tasks and in uprade steps where we really want to make sure that we handle all the objects, no matter what the security is.
(And here we actually are fixing the security, so we shouldn't rely on it being correct 😜 )

lukasgraf: Good point, thanks for the hint!
But don't you mean **un**restrictedTraverse? `restrictedTraverse` will still check for security. Also, using [`brain._unrestrictedGetObject()`](https://github.com/zopefoundation/Products.ZCatalog/blob/2a21568ebc0f489d9cda3109365015fe17d8efa3/src/Products/ZCatalog/CatalogBrains.py#L51-L64) would be another option, closer do what we actually want to do here.

jone: You're right; I've fixed restricted => unrestricted 😉

PR #120, opened 2018-04-06
Add script to delete orphaned OGDS tasks.

lukasgraf:
This script **DELETES** orphaned OGDS tasks.

OGDS tasks are considered orphaned if no Plone task exists at the `physical_path` that the OGDS record points to.

These can occur as a leftover if the zope instance was killed during task creation for some reason.

Usage:

```
    bin/instance run delete_orphaned_ogds_tasks.py --i-know-what-im-doing -n
```

The script has several safeguards. Before finally considering a task orphaned, it checks that
- we can traverse to the physical path of the container that the task supposedly is in
- the physical path doesn't appear in the Plone catalog
- the IntID doesn't resolve to a Plone object
- the task is neither a subtask nor does it have a predecessor

### DIFF
--- a/opengever/maintenance/scripts/delete_orphaned_ogds_tasks.py
+++ b/opengever/maintenance/scripts/delete_orphaned_ogds_tasks.py
@@ -1,0 +1,122 @@
+"""
+DELETES orphaned OGDS tasks.
+
+OGDS tasks are considered orphaned if no Plone task exists at the
+`physical_path` that the OGDS record points to.
+
+These can occur as a leftover if the zope instance was killed during task
+creation for some reason.
+
+    bin/instance run delete_orphaned_ogds_tasks.py --i-know-what-im-doing -n
+
+"""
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/delete_orphaned_ogds_tasks.py
```diff
@@ -0,0 +1,122 @@
...
+    current_au_id = get_current_admin_unit().unit_id
+    ogds_tasks = session.query(Task).filter_by(admin_unit_id=current_au_id)
+
+    for ogds_task in ogds_tasks:
+        assert ogds_task.admin_unit_id == plone.id
+        path = str(ogds_task.physical_path)
+        container_path = '/'.join(path.split('/')[:-1])
+
+        # Attempt to traverse to the supposed task's physical_path. If it
+        # can't be found, it is considered orphanaed and should be safe to
```
phgross: typo: `orphanaed`

PR #134, opened 2018-06-11
Add script to check for non-persisted fields.

lukasgraf:
This script is an expansion of the one being removed (introduced in #133).

It now:
- Checks *all* objects (assuming they're reachable via catalog)
- Builds the lists of expected schemas and fields dynamically (and caches them)
- Produces output in CSV format on STDOUT (for easier processing), and some progress info on STDERR


lukasgraf:
@deiferni assigned you directly because you already had a look at the previous one :)

PR #144, opened 2018-09-11
Add script to fix non-persisted values.

lukasgraf:
This script **persists field values on all objects**. It does so by querying the catalog for all objects, and persisting any field values that currently aren't persisted (=volatile) by
  - iterating over all of the schemas of the object's portal_type
  - iterating over every field of each schema
  - determining the value that the field should have

Determining the correct value is done using a few different strategies:
- Fields which are **optional** and **don't have any default** are whitelisted as such. These fields get their **missing_value** persisted.
- Fields which are **optional** and have a **static default** (i.e. no `defaultFactory`) are whitelisted as such. These fields get their **static default** persisted.
- For more complex cases, some **custom handlers** are used to deal with specific fields individually

For any left over fields that haven't been handled explicitly, and exception is raised and the script aborts.

Objects are reindexed as necessary. The script tries its best to efficently determine whether reindexing of metadata or any indexes is required, but errs on the side of caution.

| Deployment | Duration | Fixed Objects | Total Objects | % affected | time per checked obj | DB size growth |
| --- | --- | --- | --- | --- | --- | --- |
| dev-fd | 1m50.015s | 1868 | 3162 | 59.1% | .027s | 0.36% |
| kgs | 4m33.637s | 9880 | 17877 | 55.3% | .014s | 3.8% |

### DIFF
--- a/opengever/maintenance/scripts/fix_non_persisted_values.py
+++ b/opengever/maintenance/scripts/fix_non_persisted_values.py
@@ -1,0 +1,952 @@
+"""
+Script to fix (persist) field values that haven't been persisted on objects.
+
+    bin/instance run fix_non_persisted_values.py
+
+This script logs a detailed CSV report and a summary to var/log/, and displays
+some progress info and stats on STDERR/STDOUT.
+"""
+
+from App.config import getConfiguration
+from collections import Counter
+from collections import namedtuple
+from datetime import datetime
+from datetime import timedelta
+from opengever.base.default_values import get_persisted_value_for_field
+from opengever.dossier.dossiertemplate.behaviors import IDossierTemplateSchema
+from opengever.maintenance.debughelpers import setup_app
+from opengever.maintenance.debughelpers import setup_option_parser
+from opengever.maintenance.debughelpers import setup_plone
+from opengever.meeting.interfaces import IMeetingDossier
+from opengever.meeting.proposal import IProposal
+from opengever.task.task import ITask
+from operator import itemgetter
+from plone import api
+from plone.dexterity.utils import iterSchemataForType
+from plone.indexer.interfaces import IIndexableObject
+from plone.indexer.interfaces import IIndexer
+from Products.PluginIndexes.DateIndex.DateIndex import DateIndex
+from zope.component import getUtility
+from zope.component import queryMultiAdapter
+from zope.intid.interfaces import IIntIds
+from zope.schema import getFieldsInOrder
+import logging
+import os
+import sys
+import transaction
+
+
+# Lightweight data structure to keep track of field values that got persisted
+FixedField = namedtuple(
+    'FixedField', [
+        'schema_name',    # (short) name of the field's schema interface
+        'field_name',     # name of the field
+        'new_value',      # repr() of the value that got persisted
+        'value_changed',  # boolean indicating whether the value changed or
+                          #   not (compared to the volatile value)
+    ]
+)
+
+# Fields that are not required, and have a field.default (but *not*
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/fix_non_persisted_values.py
```diff
@@ -0,0 +1,955 @@
...
+
+
+class Reindexer(object):
+
+    def __init__(self, fixer):
+        self.fixer = fixer
+        self.catalog = fixer.catalog
+        self.stats = fixer.stats
+
+        self.fixer = None
```
deiferni: this line is weird, it sets `fixer` to `None` but it is just assigned above. `self.fixer` does not seem to be read, maybe the two assignments are no longer necessary?

opengever/maintenance/scripts/fix_non_persisted_values.py
```diff
@@ -0,0 +1,955 @@
...
+
+        self.stats['update_metadata'][update_metadata] += 1
+
+        idxs_needing_reindex = self.get_idxs_needing_reindex(
+            obj, fixed_fields, brain)
+
+        if update_metadata or idxs_needing_reindex:
+            # If idxs == [] the catalog defaults to *all* indexes, so we
+            # supply it a cheap index to trick it into only rebuilding metadata
+            if idxs_needing_reindex == []:
```
deiferni: is there a reason to to an explicit comparison to an empty list 🤔? i would have expected:

```python
if not idxs_needing_reindex:
```

lukasgraf:
updated...

opengever/maintenance/scripts/fix_non_persisted_values.py
```diff
@@ -0,0 +1,952 @@
...
+        if update_metadata or idxs_needing_reindex:
+            # If idxs == [] the catalog defaults to *all* indexes, so we
+            # supply it a cheap index to trick it into only rebuilding metadata
+            if not idxs_needing_reindex:
+                idxs_needing_reindex = ['getId']
+
+            sys.stderr.write(
+                "Reindexing %s (update_metadata=%r, idxs=%r)\n" % (
+                    obj, update_metadata, idxs_needing_reindex))
+            self.catalog.reindexObject(obj, idxs=idxs_needing_reindex,
```
phgross: For SOLR deployments we would need to do a manual solr-reindex here, because solr reindex is not triggered when using `catalog.reindexObject`.

opengever/maintenance/scripts/fix_non_persisted_values.py
```diff
@@ -0,0 +1,952 @@
...
+            self.log("Field %r has a defaultFactory and no custom handler, "
+                     "refusing to persist its value" % fieldname)
+            try:
+                val = get_persisted_value_for_field(obj, field)
+                self.log("Currently persisted value: %r" % val)
+            except AttributeError:
+                self.log("Currently persisted value: <NO PERSISTED VALUE>")
+
+            self.log("")
+            raise Exception(
```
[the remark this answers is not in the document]
lukasgraf: @phgross as discussed: I will check ZG special dossiers before we run the script there, instead of treating them ahead of time.

opengever/maintenance/scripts/fix_non_persisted_values.py
lukasgraf: @phgross added support for Solr indexing. This currently works for deployments that have Solr enabled and ones that don't - but for older GEVER releases that don't even have `ftw.solr` in their eggs there still needs to be conditional imports. But I'll adapt the script once we get into that situation.

PR #149, opened 2018-10-16
Add script to check object conformance to schema.

njohner:
We add a script which checks all objects for conformance to their schemata. The script outputs statistics to a STDERR/STDOUT as well as to a log file and writes a detailed CSV report.

We output 3 different types of statistics:

**Global statistics** showing a summary of the number of objects of each types and how many are conforming or not to their schemata.
screen shot 2018-10-16 at 14 53 50

**Detailed statistics** For each `portal_type` we present two tables. 
* In the first one we list the problematic fields and generally what type of problem was present (missing value, invalid, or somehow validation failed).
* In the second table we list the different types of errors encountered for each field.
screen shot 2018-10-16 at 14 54 23

With the `--verbose` option, we aggregate the errors not according to error type but error message. This is useful as it shows more information, usually the value of the field. This allows to easily see problems with default values, as the exact same message will show up many times. This will probably be an overwhelming amount of output on a big installation.
screen shot 2018-10-16 at 15 12 17

This is for issue 4teamwork/opengever.core#4440. I'm not saying resolves here as I think we may want to leave that issue open?

### DIFF
--- a/opengever/maintenance/scripts/validate_object_schema_conformance.py
+++ b/opengever/maintenance/scripts/validate_object_schema_conformance.py
@@ -33,6 +33,54 @@ some progress info and stats on STDERR/STDOUT.
 """
 
 
+class TextTable(object):
+
+    def __init__(self, column_definitions=None, separator=" | ", with_title=True):
+        self.column_definitions = column_definitions
+        self.separator = separator
+        self.data = []
+        self.with_title = with_title
+
+    def add_row(self, row):
+        self.data.append(row)
+
+    @property
+    def ncols(self):
+        if len(self.data) == 0:
+            return 0
+        return len(self.data[0])
+
+    def calculate_column_width(self):
+
+        self.widths = [0 for i in range(self.ncols)]
+        for row in self.data:
+            for i, el in enumerate(row):
+                if len(str(el)) > self.widths[i]:
+                    self.widths[i] = len(str(el))
+
+    def get_format_string(self):
+        self.calculate_column_width()
+        frmtstr = []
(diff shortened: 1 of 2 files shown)


njohner:
@lukasgraf Don't know if this one got forgotten or is still in your stack

lukasgraf:
@njohner the former, sorry 😢 Will have a look...

opengever/maintenance/scripts/validate_object_schema_conformance.py
```diff
@@ -0,0 +1,339 @@
...
+
+        for schema in schemas:
+            if schema.__identifier__ not in self.FIELD_CACHE:
+                self.FIELD_CACHE[schema.__identifier__] = getFieldsInOrder(schema)
+            fields = self.FIELD_CACHE[schema.__identifier__]
+
+            for name, field in fields:
+                field_instance = field.bind(obj)
+                try:
+                    value = field_instance.get(obj)
```
lukasgraf: This should be `field_instance.get(field.interface(obj))` (object needs to be adapted to schema to get a possible storage adapter). This might possibly explain some of the high nonconformance rates, especially for the LifeCycle behavior.

njohner: Right, after correction, `LifeCycle` does not show up as much anymore, and overall now ~50% of the objects came out as conforming on my local installation (instead of the 10% measured before that fix).

njohner:
Ok I rebased and corrected the error you spotted. I also added the forgotten CL entry.
This should be ready.

PR #153, opened 2018-11-26
Add script to fix journal pdf document date and add IDossierJournalPDFMarker.

njohner:
We add a maintenance script to fix resolved dossiers in invalid states because of their journal PDF.
This can happen for two reasons:
* dossier was closed before `2018.3.0` (https://github.com/4teamwork/opengever.core/pull/4277), so that the journal PDF can have an invalid `document_date`
* dossier was closed before `2018.5.0` (https://github.com/4teamwork/opengever.core/pull/4806/files), so that the journal pdf is not marked with the `IDossierJournalPDFMarker`.

There is actually no need to correct the first of these two problems anymore, but all documents presenting this problem should also present the second one, so we just fix both.

We restrict ourselves to pdf documents in dossiers that are in the resolved state. We also filter out documents that provide the `IAutoGeneratedDocument` interface, as these will anyway be ignored in the resolve conditions. I could not do this exclusion directly in the catalog queries. It seems we do not support the `not` keyword in queries on `object_provides`.

I'm also checking that the document title matches the expected title of a dossier journal. There is nevertheless an option `-f` to correct these cases anyway.

This resolves https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/427


opengever/maintenance/scripts/fix_journal_pdf_document_date.py
```diff
@@ -0,0 +1,123 @@
...
+                brain.Title, brain.getPath())
+            alsoProvides(document, IDossierJournalPDFMarker)
+            to_index.append('object_provides')
+
+    if to_index:
+        document.reindexObject(idxs=to_index)
+
+
+def get_problematic_journal_pdfs():
+    dossier_brains = api.content.find(portal_type='opengever.dossier.businesscasedossier')
```
phgross: IMHO it should be possible to only check `resolved` dossiers, or not?

njohner:
Hum, looks like I did not push my local changes 😓 ...

njohner:
Voila, should be better now sorry. I had not committed my changes before pushing, sorry!

PR #155, opened 2018-12-04
Improve solr activation and add a solr maintenance script

phgross:
- Add solr-maintenance bin/instance run script, which allows to start solr-maintenance jobs from the shell. `reindex`, `clear`, `diff` and `sync` are the supported modes.
- Use sync mode in the solr activate script for indexing content.  So its also possible to reindex and prepare a large GEVER deployment for the solr activation and run the solr activation in a separate step.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,8 @@ Changelog
 1.0 (unreleased)
 ----------------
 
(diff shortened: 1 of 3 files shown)


phgross:
@deiferni the solr maintenance view is already an adapter 😜

deiferni:
yes as it is a view 😉

PR #160, opened 2019-01-07
Add script to fix documents incorrectly marked as journal pdf

### LINKED ISSUE #159: Remove wrongly assigned IDossierJournalPDFMarker interfaces
With the `-f` option of the `fix_journal_pdf_document_date.py` script (https://github.com/4teamwork/opengever.maintenance/pull/153), `IDossierJournalPDFMarker` was potentially added to documents that are not journal pdfs. We did not realise that there were other documents than journal PDFs that had a `changed` date or `document_date` incompatible with the dossier `end` date (see https://github.com/4teamwork/opengever.core/issues/5204).

We should:
- Remove the `-f` option of that script
- Add a script to correct the wrongly assigned interfaces

njohner:
With the `fix_journal_pdf_document_date.py` script, we potentially added the `IDossierJournalPDFMarker` interface to documents that are not journal PDFs. Here we add a script to remove this interface from documents.

We also fix the `fix_journal_pdf_document_date.py` script to avoid wrongly marking documents in the future.

Finally several utilites were moved into `utils.py` so that they can more easily be reused by other maintenance scripts.

resolves #159


njohner:
I'll feed the piggy for the shameful bug `fix_journal_pdf_document_date.py` script

lukasgraf:
@njohner not sure that's 🐷 worthy, but that's up to you 😉

njohner:
hum, I'm not sure we did run the faulty script on many systems. But I wouldn't know where we did and where we didn't...

njohner:
❗️ Do not merge ❗️ 
I had to do quite a rebase here, as I'd cherry-picked several of the commits from here to other PRs that have been merged since. I'm testing that it still runs properly :-).
Also concerning the squashing, I'm happy to squash more, but I've split the commits on purpose, because the changes on `fix_journal_pdf_document_date.py` are actually rather small, but get lost in the refactoring.

deiferni:
sorry about the squashing, i misinterpreted. you amend to an already existing script, not the one you add in this script. that is very helpful 😄.

PR #162, opened 2019-01-10
Add script to list proposals with no returned excerpt.

njohner:
We add a script that lists the decided proposals in closed meetings, but lacking a returned excerpt. As it is now mandatory to return an excerpt for each proposal before closing a meeting, this allows us to get an idea of how many proposals are not conforming to this new rule.

I also added the possibility to wrap text in the `TextTable` class, as to make the output more easily readable in the console. Drawback is that path might span several lines and be harder to copy and paste...


opengever/maintenance/scripts/list_decided_proposals_without_excerpt.py
```diff
@@ -0,0 +1,72 @@
...
+
+class DecidedProposalWithoutExcerptLister(object):
+
+    def __init__(self):
+        self.table = TextTable(col_max_width=60)
+        self.table.add_row(["proposal path", "proposal title"])
+
+    def list_proposals_without_excerpt(self):
+        for proposal in self.get_proposals_without_excerpt():
+            proposal.agenda_item
```
deiferni: these statements do nothing and seem like a leftover.

deiferni:
oh, i removed closing the og.core issue from the description. we still have to run it for our customers and make sure the data is ok (the second item in the list).

opengever/maintenance/scripts/list_decided_proposals_without_excerpt.py
njohner: Oh I'm sorry, that looks like a dirty leftover...

njohner:
Merge base is not master on purpose. I use utilities here that are added in the other branch. PR is pending there and has not been merged yet.
But I actually cherry-picked the utility things into https://github.com/4teamwork/opengever.maintenance/pull/163, which I'll merge this morning. So I'll rebase this branch onto master then

njohner:
Ok I rebased this on master as the commits I needed were merged into master in https://github.com/4teamwork/opengever.maintenance/pull/163.
I also included your comments. This is ready to merge

PR #163, opened 2019-01-22
Add script to fix filenames with missing extension

njohner:
We add a script to find all documents and mails with missing filename extension and add the extension found in the mimetype registry.

In this PR we also:
* move `TextTable` into `utils.py` so that it can more easily be reused by other maintenance scripts
* Added `LogFilePathFinder` class to `utils.py`, mainly copied from other maintenance scripts (https://github.com/4teamwork/opengever.maintenance/blob/master/opengever/maintenance/scripts/find_non_persisted_values.py). This was also used in several script and provides a good way of having uniform naming conventions for log files produced by maintenance scripts.

resolves https://extranet.4teamwork.ch/support/ph-vs/tracker/534


opengever/maintenance/scripts/add_missing_file_extensions.py
```diff
@@ -0,0 +1,123 @@
...
+        mimetype = None
+        try:
+            mimetype = document.get_mimetype()[0]
+            extension = self.registry.lookup(mimetype)[0].extensions[0]
+            new_filename = u"{}.{}".format(filename, extension)
+            document.get_file().filename = new_filename
+            self.corrected.add_row((path, filename, mimetype, new_filename))
+        except:
+            self.skipped.add_row((path, filename, mimetype))
+
```
njohner: Not sure whether it is acceptable here to use a bare `except`. I could obviously handle all that a bit more cleanly, but for a maintenance script I thought that would do, especially since we list the filenames that could not be corrected.

njohner:
I added the print out of the exception. I'll merge the commits when you're done reviewing

opengever/maintenance/scripts/add_missing_file_extensions.py
lukasgraf: As discussed, I would probably just `print` the exception (plus some context, like `path`)- but other than that it's totally fine for a maintenance script. The key with no using bare `except:`'s is that exceptions might go unhandled. But appending it to `self.skipped` (and printing) *is* actually handling it in some way.

PR #170, opened 2019-03-20
Add script to list all dossiers shared with a group.

njohner:
More specifically the scripts finds all dossiers on which the group is directly assigned local roles through sharing. This is for https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/605

### DIFF
--- a/opengever/maintenance/scripts/list_dossiers_shared_with_group.py
+++ b/opengever/maintenance/scripts/list_dossiers_shared_with_group.py
@@ -1,0 +1,94 @@
+"""
+Searches for all resolved dossiers, on which local_roles were set for a given
+group via sharing.
+
+    bin/instance run ./scripts/list_dossiers_shared_with_group.py group_name
+
+mandatory arguments:
+  group_name : name of the group for which we will search dossiers.
+
+"""
+from opengever.base.role_assignments import ASSIGNMENT_VIA_SHARING
+from opengever.base.role_assignments import RoleAssignmentManager
+from opengever.maintenance.debughelpers import setup_app
+from opengever.maintenance.debughelpers import setup_option_parser
+from opengever.maintenance.debughelpers import setup_plone
+from opengever.maintenance.utils import LogFilePathFinder
+from opengever.maintenance.utils import TextTable
+from plone import api
+import sys
+
+
+class DossierSharedWithGroupLister(object):
+
+    def __init__(self, group):
+        self.table = TextTable(col_max_width=60)
+        self.table.add_row(["dossier path", "dossier title", "roles"])
+        self.group = group
+
+    def list_dossiers_shared_with_group(self):
+        for dossier, assignment in self.get_dossiers_shared_with_group():
+            self.table.add_row([dossier.absolute_url_path(), dossier.Title(), assignment.roles])
+
+    def print_table(self):
+        print("Table of dossiers shared with {}".format(self.group.id))
+        print(self.table.generate_output())
+        print("\nSummary:")
+        print("There are {} dossiers shared with {}".format(self.table.nrows,
+                                                            self.group.id))
+
+    def get_dossiers_shared_with_group(self):
+        """ Searches for all dossier shared with a given group.
+        We make the assumption that the group has at least view permissions
+        on such dossiers.
+        """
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/list_dossiers_shared_with_group.py
```diff
@@ -0,0 +1,92 @@
...
+        on such dossiers.
+        """
+        dossier_brains = api.content.find(allowedRolesAndUsers=u'user:{}'.format(self.group.id))
+        for dossier_brain in dossier_brains:
+            dossier = dossier_brain.getObject()
+            assignments = RoleAssignmentManager(dossier).get_assignments_by_principal_id(self.group.id)
+            sharing_assignment = self._find_sharing_assignment(assignments)
+            if sharing_assignment:
+                yield (dossier, sharing_assignment)
+        return
```
deiferni: As this is a generator i'd omit the `return` in combination with `yield`. I'm surprised this even works 🤔.

opengever/maintenance/scripts/list_dossiers_shared_with_group.py
```diff
@@ -0,0 +1,92 @@
...
+
+    @staticmethod
+    def _find_sharing_assignment(assignments):
+        try:
+            index = map(lambda assignment: assignment.cause, assignments).index(ASSIGNMENT_VIA_SHARING)
```
deiferni: i find this much harder to read than something like

```python
sharing_assignments = [each for each in assignments if each.cause == ASSIGNMENT_VIA_SHARING]
```

also, could there be multiple assignments? probably not, right 🤔?

opengever/maintenance/scripts/list_dossiers_shared_with_group.py
```diff
@@ -0,0 +1,92 @@
+"""
+Searches for all resolved dossiers, on which local_roles were set for a given
+group via sharing.
+
+    bin/instance run ./scripts/list_dossiers_shared_with_group.py
```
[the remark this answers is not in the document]
njohner: Yes sure, copy/paste fail...

opengever/maintenance/scripts/list_dossiers_shared_with_group.py
njohner: I think that is how generators work. As long as they yield something it goes on, when they return `None`, it's done

njohner: Of course I can remove the return. But I think it's implicit anyway

deiferni: Mhmm. Just be aware that the `yield` implicitly converts the method into a generator. which behaves slightly different. 

Something like
```python
def foo():
    for i in range(3):
      yield i
    return None      
```
will cause a `SyntaxError` in python 2.

I think https://dzone.com/articles/on-generator-functions-yield-and-return explains it a bit more in depth, in case you got time to do some reading.

So a simple `return` will work as that seems to be treated the same as raising `StopIteration`  (didn't know that before 😄), but you can't return an actual value as the function is a generator and thus must `yield`.

opengever/maintenance/scripts/list_dossiers_shared_with_group.py
njohner: yes there could be multiple assignments. Maybe one from sharing and one from a forwarding for example

njohner: But only one from assignment. That's why I look for the index.

opengever/maintenance/scripts/list_dossiers_shared_with_group.py
njohner: Ok thanks!

opengever/maintenance/scripts/list_dossiers_shared_with_group.py
deiferni: yes per category and user i meant. but don't see how either.

opengever/maintenance/scripts/list_dossiers_shared_with_group.py
njohner: I read and I learnt...

PR #177, opened 2019-06-07
Fix value changed reporting, respect all fields.

deiferni:
Report if the value has changed in any field, not just the last field from the scope of the previous list comprehension.


deiferni:
Successful testrun on a larger test deployment, the values seem to be reported correctly now.

PR #181, opened 2019-07-19
Add nightly job and script to queue missing archival file conversions.

lukasgraf:
This PR adds a script that produces
- a **report of documents missing their archival file**
- optionally **queues them for conversion**

and 
- a nightly job provider that **triggers conversion for those queued documents**.

> **Usage**: archival_file_checker.py [-n] [report_missing | queue_missing]
>
> The script takes one of two commmands:
>
> `report_missing`
>     Reports all documents missing their archival file.
>
> `queue_missing`
>     Reports the documents, and additionally queues them for conversion by
>     a nightly job (by adding them to a persistent queue on the site root).
>
> The script will display some basic console output, and automatically log
> that ouput and some more detailed information to a logfile.

### Identifying documents missing their archival file 

Identifying the documents that should have an archival file, but don't, is done as follows:
For all **candidate dossiers**, check if the documents contained in them are missing their archival file.

A candidate dossier is any dossier that:
- is resolved
- doesn't have any after-resolve jobs pending
- isn't a subdossier (we'll check docs recursively)

For each document in those dossiers we then determine if it acutally _needs_ an archival file. `IBaseDocument`s that don't are:
- Mails
- Dossier Journal PDFs
- Documents that aren't Bumblebee-convertable

If the document should have an archival file but doesn't, it's flagged as `missing`.

For each dossier we'll gather some stats that allow us to cross-check that the script is operating correctly, even though in the end we only really need a list of documents (per dossier) that are missing their archival file.

Once done, we'll display (and log) a summary report. In addition, a detailed list of the documents missing their archival file is logged to the file (but not displayed).

We'll also show the current queue length at the end.

### Queueing documents for conversion of their missing archival file

The documents for which we'll want to trigger conversion are written to a persistent queue when the script is invoked with the `queue_missing` command. This queue is an IOBTree on the site root annotations - a mapping from a dossier's IntId to a list of document IntIds in that dossier that need conversion.

It is grouped by dossier because this allows us to process the queue in more reasonably sized chunks, instead of every single document being considered a "job" (which would lead to a commit every time).

So the unit of work for the nightly job is a dossier, and during the execution of that job it will trigger the conversion for that dossier's missing archival files.

### Nightly job to convert documents from the queue

A nightly job will then present the dossier IntIds (the keys in the queue) to the runner as it's list of jobs, and, when invoked with such a dossier IntId, trigger the conversion of the document IntIds that are listed in the queue under that dossier IntId.

The nightly job includes two measures to try to not overload Bumblebee too much, and possibly displace other, more important conversion jobs (nightly dossier resolution, or preview PDFs during the day):
- Each conversion is followed by a `time.sleep(1)` to at least stagger the conversion requests a little bit
- There is a `MAX_CONVERSION_REQUESTS_PER_NIGHT` (currently 1000). Once that many conversion requests have been queued by the archival file job during one nightly run, it will cease its work and further processing of the queue will resume the following night.

### DIFF
--- a/opengever/maintenance/scripts/archival_file_checker.py
+++ b/opengever/maintenance/scripts/archival_file_checker.py
@@ -1,0 +1,368 @@
+"""
+Script that reports documents missing their archival file, and optionally
+queues them for conversion by a nightly job.
+
+Usage: archival_file_checker.py [-n] [report_missing | queue_missing]
+
+The script takes one of two commmands:
+
+report_missing
+    Reports all documents missing their archival file.
+
+queue_missing
+    Reports the documents, and additionally queues them for conversion by
+    a nightly job (by adding them to a persistent queue on the site root).
+
+The script will display some basic console output, and automatically log
+that ouput and some more detailed information to a logfile.
+"""
+from BTrees.IIBTree import IITreeSet
+from BTrees.IOBTree import IOBTree
+from collections import Counter
+from collections import OrderedDict
+from ftw.bumblebee.interfaces import IBumblebeeDocument
+from opengever.document.behaviors import IBaseDocument
+from opengever.dossier.behaviors.dossier import IDossierMarker
+from opengever.maintenance.debughelpers import setup_app
+from opengever.maintenance.debughelpers import setup_plone
+from opengever.maintenance.nightly_archival_file_job import MISSING_ARCHIVAL_FILE_KEY
+from opengever.maintenance.utils import LogFilePathFinder
+from opengever.maintenance.utils import TextTable
+from plone import api
+from Products.CMFPlone.interfaces import IPloneSiteRoot
+from zope.annotation import IAnnotations
+from zope.component import getUtility
+from zope.intid.interfaces import IIntIds
+import argparse
+import logging
+import sys
+import transaction
+
+
+logger = logging.getLogger('archival_file_checker')
+logger.setLevel(logging.INFO)
+
+
+class ArchivalFileChecker(object):
+    """Checks whether documents that should have an archival file actually
+    do have one, and reports the ones that don't.
+
+    With command 'queue_missing' also queues those documents for conversion
+    via a nightly job.
+    """
+
+    def __init__(self, context, options, logger):
+        self.context = context
+        self.options = options
+        self.logger = logger
+
+        self.log = logger.log
+        self.log_to_file = logger.log_to_file
+
+        self.catalog = api.portal.get_tool('portal_catalog')
+        self.all_dossier_stats = None
+
+    def run(self):
+        assert IPloneSiteRoot.providedBy(self.context)
+
+        missing_by_dossier = self.check()
+
+        if self.options.cmd == 'queue_missing':
+            self.queue_missing(missing_by_dossier)
+
+        self.log("")
+        self.log("Detailed log written to %s" % self.logger.logfile_path)
+
+    def check(self):
+        """For all candidate dossiers, check if the documents contained in
+        them are missing their archival file.
+
+        A candidate dossier is any dossier that:
+        - is resolved
+        - doesn't have any after-resolve jobs pending
+        - isn't a subdossier (we'll check docs recursively)
+
+        For each dossier we'll gather some stats that allow us to cross-check
+        that the script is operating correctly, even though in the end we
+        only really need a list of documents (per dossier) that are missing
+        their archival file.
+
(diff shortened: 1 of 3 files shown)


opengever/maintenance/scripts/archival_file_checker.py
```diff
@@ -0,0 +1,394 @@
...
+                if should_have_archival_file:
+                    dossier_stats['missing'] += 1
+                    docs_missing_archival_file.append(doc)
+
+        return dossier_stats, docs_missing_archival_file
+
+    def should_have_archival_file(self, doc):
+        """Determine whether an IBaseDocument should have an archival file.
+        """
+        if doc.portal_type == 'ftw.mail.mail':
```
njohner: `mails` have an `is_mail` attribute

opengever/maintenance/scripts/archival_file_checker.py
```diff
@@ -0,0 +1,394 @@
...
+        """Queue conversion job for documents missing their archival file.
+
+        This method takes a list of dicts, one per dossier that contains at
+        least one document with a missing archival file.
+        """
+        self.log("")
+        self.log("Queueing missing archival files")
+        self.log("=" * 80)
+
+        total_missing = sum([len(g['missing']) for g in missing_by_dossier])
```
njohner: `group` instead of `g`

opengever/maintenance/scripts/archival_file_checker.py
```diff
@@ -0,0 +1,394 @@
...
+
+        for group in missing_by_dossier:
+            dossier = group['dossier']
+            missing = group['missing']
+            print "  Queueing %s documents for %r" % (len(missing), dossier)
+
+            intids = getUtility(IIntIds)
+            dossier_intid = intids.getId(dossier)
+
+            queue[dossier_intid] = IITreeSet()
```
njohner: So if the queue already existed, we overwrite entries for dossiers that were already in the queue, but would leave entries that are in the queue but not found anymore by this run (for some reason or another). Maybe add a message in the log when `dossier_intid` is already in the queue?

opengever/maintenance/scripts/archival_file_checker.py
```diff
@@ -0,0 +1,394 @@
...
+        self.logger = logger
+
+        self.log = logger.log
+        self.log_to_file = logger.log_to_file
+
+        self.catalog = api.portal.get_tool('portal_catalog')
+        self.all_dossier_stats = None
+
+    def run(self):
+        missing_by_dossier = self.check()
```
njohner: Basically `run` will fail if the context is not the site root. But it will fail only after doing all the work. I'd add the check `assert IPloneSiteRoot.providedBy(self.context)` wither here before running the `self.check()` or even in the `__init__`

opengever/maintenance/scripts/archival_file_checker.py
```diff
@@ -0,0 +1,394 @@
...
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.logfile.close()
+
+    def get_logfile_path(self, filename):
+        log_dir = self.get_logdir()
+        return os.path.join(log_dir, filename)
+
+    def get_logdir(self):
```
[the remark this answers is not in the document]
lukasgraf: Nice! Makes it much cleaner 😄 💚

opengever/maintenance/scripts/archival_file_checker.py
```diff
@@ -0,0 +1,394 @@
...
+        for doc_brain in contained_docs:
+            doc = doc_brain.getObject()
+
+            # Determine if this document should have an archival file
+            should_have_archival_file = self.should_have_archival_file(doc)
+
+            if should_have_archival_file:
+                dossier_stats['should_have_archival_file'] += 1
+
+            # Check if an archival file is present
```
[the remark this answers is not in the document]
lukasgraf: As discussed: If we wouldn't do this, the stats would be off in the case where we encounter spurious archival files. And the only reason for the stats being there anyway is to cross-check the script's results. So for that reason I'm leaving it as it is.

opengever/maintenance/scripts/archival_file_checker.py
lukasgraf: Yes, indeed. As suggested, I added logging for the case when we're overwriting/replacing existing queue entries. (Also changed some `print` statements to `self.log` that I seem to have missed 😉 )

PR #182, opened 2019-07-22
ArchivalFileChecker: Correctly handle pending resolve jobs.

lukasgraf:
Querying the catalog for `after_resolve_jobs_pending=False` does not include objects with `Missing.Value`, only ones that actually are indexed with `False`.

We therefore need to check on the actual objects for pending after resolve jobs instead of relying on the catalog.


lukasgraf:
Already tested like this on PROD :trollface: , I'm therefore self-merging this.

PR #184, opened 2019-07-25
Add InboxGroupMigrator script to solve support issue after inbox_group change.

phgross:
https://extranet.4teamwork.ch/support/stadt-nidau/tracker-gever/140

We were discussing whether we should implement the whole thing as an upgrade step. I decided against it and implemented a parameterizable script instead. Which should be good for now, moving this to opengever.core could be part of the `Personalmutation OGIP`. I made me a note.

See https://github.com/4teamwork/opengever.nidau/issues/35

### DIFF
--- a/opengever/maintenance/scripts/update_inbox_group.py
+++ b/opengever/maintenance/scripts/update_inbox_group.py
@@ -1,0 +1,96 @@
+"""A script to migrate role assignments of tasks and related documents and
+the distinct parent, after changing the inbox_group of an org_unit.
+
+See https://extranet.4teamwork.ch/support/stadt-nidau/tracker-gever/140
+
+bin/instance run src/opengever.maintenance/opengever/maintenance/scripts/update_inbox_group.py 'sd' 'GG_NID_SD' 'GG_NID_SD_Bereichsleiter'
+"""
+
+from opengever.base.oguid import Oguid
+from opengever.base.role_assignments import ASSIGNMENT_VIA_TASK
+from opengever.base.role_assignments import ASSIGNMENT_VIA_TASK_AGENCY
+from opengever.base.role_assignments import RoleAssignmentManager
+from opengever.maintenance.debughelpers import setup_app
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/update_inbox_group.py
```diff
@@ -0,0 +1,99 @@
+# A script to migrate role assignments of tasks and related documents and
```
deiferni: this looks like a module level docstring thus it should be enclosed by `"""` not by single line comments.

opengever/maintenance/scripts/update_inbox_group.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def migrate(self):
+        catalog = api.portal.get_tool('portal_catalog')
+        brains = catalog.unrestrictedSearchResults(
+            object_provides=ITask.__identifier__)
+
+        org_units_tasks = []
+        for brain in brains:
+            task = brain.getObject()
+            if task.responsible_client == self.org_unit:
```
deiferni: couldn't you use the `responsible_client` index?

opengever/maintenance/scripts/update_inbox_group.py
```diff
@@ -0,0 +1,99 @@
...
+
+        for assignment in assignments:
+            if reference_oguid == assignment.reference:
+                continue
+
+            if assignment.cause in [ASSIGNMENT_VIA_TASK, ASSIGNMENT_VIA_TASK_AGENCY]:
+                reference = Oguid.parse(assignment.reference).resolve_object()
+                manager.add_or_update(self.new_inbox_group, assignment.roles,
+                                      assignment.cause, reference)
+                manager.clear(assignment.cause, assignment.principal, reference)
```
deiferni: Could we use `self.old_inbox_group` as the principal passed to `clear` to clarify what is going on? Should we first `clear` the old assignment and then add the new one?

phgross:
@deiferni updated according to your feedback ...

PR #187, opened 2019-08-09
Add fallback for konsulmigration relicts to the persist defaultvalues script.

phgross:
According to the discussion we had in https://github.com/4teamwork/opengever.zug/issues/315#issuecomment-518150992.


opengever/maintenance/scripts/fix_non_persisted_values.py
```diff
@@ -656,7 +657,21 @@ def _get_volatile_value(self, obj, field):
...
 
         # Verify that the value is valid
-        bound_field.validate(volatile_value)
+        try:
+            bound_field.validate(volatile_value)
+        except RequiredMissing as e:
+            # Check if its a part of the `Abnahme` repositoryfolder,
+            # a relict of the konsulmigration (content which was not
+            # created correctly). In that case we try to use the field default.
+            bound_field = field.bind(obj)
```
lukasgraf: This is already being done above, on line 655

PR #188, opened 2019-09-04
Update script to list dossiers shared with group

njohner:
Fix output and add a force option to allow listing the dossiers for a group that does not exist anymore.

### DIFF
--- a/opengever/maintenance/scripts/list_dossiers_shared_with_group.py
+++ b/opengever/maintenance/scripts/list_dossiers_shared_with_group.py
@@ -21,31 +21,36 @@ import sys
 
 class DossierSharedWithGroupLister(object):
 
-    def __init__(self, group):
+    def __init__(self, group_id):
         self.table = TextTable(col_max_width=60)
         self.table.add_row(["dossier path", "dossier title", "roles"])
-        self.group = group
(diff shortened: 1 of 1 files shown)


njohner:
It seems I forgot to get this one merged, but I have been using it for a while. This is just a small fix to allow to search for groups that do not exist anymore

opengever/maintenance/scripts/list_dossiers_shared_with_group.py
```diff
@@ -72,18 +80,18 @@ def main():
...
         sys.exit(1)
 
-    dossier_lister = DossierSharedWithGroupLister(group)
+    dossier_lister = DossierSharedWithGroupLister(groupname)
     dossier_lister.list_dossiers_shared_with_group()
     dossier_lister.print_table()
 
     log_filename = LogFilePathFinder().get_logfile_path(
-        'list_dossiers_shared_with_{}'.format(group), extension="csv")
+        'list_dossiers_shared_with_{}'.format(groupname), extension="csv")
```
deiferni: we could consistently name the variable accross the script. so use `group_id` instead of `groupname` here, as well, as i think it is the better name.

PR #199, opened 2019-12-06
Add maintenance script to reopen dossiers and close them at a later time.

phgross:
Used for https://extranet.4teamwork.ch/support/gever-st-gallen/tracker-ris/12

The script provides two modes:
**`reopen`** reopens all resolved dossiers with a start date later than the specified with the -d parameter and stores a list of the dossier uids in the annotations of the plone site.
**`closed`** try to resolve the dossier, stored in the annotations list (see above) 

The script only works for deployments with a `lenient` resolver, because the use of `getFolderContents` in the all supplied check, does not work in bin/instance run scripts. Will open up separate issue for that.

### DIFF
--- a/opengever/maintenance/scripts/reopen_and_close_rr_dossiers.py
+++ b/opengever/maintenance/scripts/reopen_and_close_rr_dossiers.py
@@ -1,0 +1,148 @@
+"""
+Script to automatically reopen dossiers and complete them at a later time.
+"""
+from AccessControl.SecurityManagement import getSecurityManager
+from AccessControl.SecurityManagement import newSecurityManager
+from AccessControl.SecurityManagement import setSecurityManager
+from contextlib import contextmanager
+from datetime import datetime
+from opengever.base.security import UnrestrictedUser
+from opengever.dossier.behaviors.dossier import IDossierMarker
+from opengever.dossier.exceptions import PreconditionsViolated
+from opengever.dossier.reactivate import Reactivator
+from opengever.dossier.resolve import LockingResolveManager
+from opengever.maintenance.debughelpers import setup_app
+from opengever.maintenance.debughelpers import setup_plone
+from persistent.list import PersistentList
+from plone import api
+from plone.app.uuid.utils import uuidToObject
+from zope.annotation import IAnnotations
+from zope.globalrequest import getRequest
+import argparse
+import logging
+import sys
+import transaction
+
+
+INPUT_DATEFMT = "%d-%m-%Y"
+ANNOTATIONS_KEY = 'rr-dossier-migration'
+
+
+logger = logging.getLogger('rr-dossier-migration')
+logging.getLogger().setLevel(logging.INFO)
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/reopen_and_close_rr_dossiers.py
```diff
@@ -0,0 +1,144 @@
...
+logger = logging.getLogger('rr-dossier-migration')
+logging.getLogger().setLevel(logging.INFO)
+for handler in logging.getLogger().handlers:
+    handler.setLevel(logging.INFO)
+
+
+@contextmanager
+def elevated_privileges(user_id=None):
+    """Copy of elevated_privileges (opengever.base.security) but also
+    assign the `Publisher` role, because re
```
njohner: Seems the end of the comment is missing

opengever/maintenance/scripts/reopen_and_close_rr_dossiers.py
```diff
@@ -0,0 +1,144 @@
...
+        tmp_user = tmp_user.__of__(api.portal.get().acl_users)
+        newSecurityManager(getRequest(), tmp_user)
+
+        yield
+    finally:
+        # Restore the old security manager
+        setSecurityManager(old_manager)
+
+
+class RRDossierManager(object):
```
njohner: Why `RR` in the class name?

opengever/maintenance/scripts/reopen_and_close_rr_dossiers.py
```diff
@@ -0,0 +1,144 @@
...
+            uids.append(brain.UID)
+            logger.info('{} reopend'.format(brain.getURL()))
+
+        self.write_list_to_the_annotiations(uids)
+        logger.info('{} dossiers reopend'.format(len(uids)))
+
+    def close_dossiers(self):
+        ann = IAnnotations(api.portal.get())
+        if ANNOTATIONS_KEY not in ann:
+            raise Exception('No dossiers reopened dossiers stored - nothing to do.')
```
njohner: No reopened dossiers stored - nothing to do.

opengever/maintenance/scripts/reopen_and_close_rr_dossiers.py
```diff
@@ -0,0 +1,144 @@
...
+    setup_plone(app, options)
+
+    if options.dryrun:
+        transaction.doom()
+
+    if options.mode == 'reopen':
+        if not options.start_date:
+            raise Exception('start_date is required in reopen mode')
+
+        start_date = datetime.strptime(options.start_date, INPUT_DATEFMT)
```
njohner: Parsing a date input from command line seems a bit prone to errors, so giving some feedback on how it was parsed would be nice. So maybe log the start_date, or print it.

phgross:
@njohner because the it only queries the catalog, the feedback-time in the dry mode should be ok.

PR #201, opened 2020-01-27
Add script to update object IDs (e.g. URLs after changing title)

lukasgraf:
This script allows to update the ID (which appears as part of the URL) of one or more objects.

Given a file with a list of paths, it will check for each of these objects what its ID would be (if it were newly created with that title), and rename it if necessary.

Usage:
```
bin/instance run update_object_ids.py [-n] <file_with_paths>
```

#### Notes
 - IDs of parents or children won't be checked/updated - only the ID of the object
   directly addressed by the given path. So this script is **NOT recursive**.
 - **Journaling will be disabled for any renames**. Therefore there won't be any journal entries for this operation.
- Necessary reindexing (for example of the `path`) will be propagated down the subtree however. So all objects below should be reindexed properly. This currently is not optimized at all in any way, so when executing it for multiple overlapping paths, it will be **quite inefficient**. Because the case in the customer's request only concerns one single repofolder, this should be ok for now.

### DIFF
--- a/opengever/maintenance/scripts/update_object_ids.py
+++ b/opengever/maintenance/scripts/update_object_ids.py
@@ -1,0 +1,153 @@
+"""
+Script to update IDs (and therefore URLs) of one or more objects.
+
+    bin/instance run update_object_ids.py [-n] <file_with_paths>
+
+<file_with_paths> should be a file with one path per line containing the list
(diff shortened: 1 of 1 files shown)


lukasgraf:
@njohner good point! I updated the script so it outright refuses to rename anything else than repository folders (it skips them).

PR #202, opened 2020-02-05
Also reindex filing_no if index is available.

deiferni:
We clear the lexicon, then reindex all remaining `ZCTextIndices`. For most deployments this is only `Title`. For some we have added the filing_no customization which adds its own `ZCTextIndex`, `searchable_filing_no`. This index also needs to be rebuilt after activating solr.

Related to https://github.com/4teamwork/opengever.core/pull/6214.

### DIFF
--- a/opengever/maintenance/scripts/activate_solr.py
+++ b/opengever/maintenance/scripts/activate_solr.py
@@ -80,13 +80,23 @@ def remove_catalog_indexes(portal):
             logger.info('Removed catalog index %s.', index)
 
     catalog.clearIndex('Title')
+    title_index = catalog._catalog.getIndex('Title')
+
+    if 'searchable_filing_no' in catalog._catalog.indexes:
+        searchable_filing_no_index = catalog._catalog.getIndex('searchable_filing_no')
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/activate_solr.py
```diff
@@ -80,13 +80,24 @@ def remove_catalog_indexes(portal):
...
 
     catalog.clearIndex('Title')
+    title_index = catalog._catalog.getIndex('Title')
+
+    if 'searchable_filing_no' in catalog._catalog.indexes:
+        searchable_filing_no_index = catalog._catalog.getIndex('searchable_filing_no')
+    else:
+        searchable_filing_no_index = None
+    if searchable_filing_no_index:
+        catalog.clearIndex('searchable_filing_no')
```
njohner: I guess you could move this up into the previous if (`if 'searchable_filing_no' in catalog._catalog.indexes:`)... but it's fine like that

deiferni: oh yes indeed, i moved things around a bit in the end and missed that. thanks!

PR #205, opened 2020-03-25
Improve memory management in list_dossiers_shared_with_group.

njohner:
For https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/965

On kapo we run out of memory when executing this script. Using commit savepoints should help manage the memory usage.


njohner:
superseded by https://github.com/4teamwork/opengever.maintenance/pull/206

PR #206, opened 2020-03-25
Improve memory management in list_dossiers_shared_with_group.

njohner:
For https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/965

On kapo we run out of memory when executing this script. Using commit savepoints should help manage the memory usage.

I copied the memory management from https://github.com/4teamwork/opengever.maintenance/pull/192, and tested on `test.onegovgever.ch`, seems to work.


njohner:
Used it and the script ran using 3.4GB, whereas before it failed after using up all the free memory (>10GB).

PR #209, opened 2020-04-07
Add script to set EMM IP ranges.

njohner:
For https://github.com/4teamwork/opengever.sg/issues/404

use with:
`bin/instance0 run src/opengever.maintenance/opengever/maintenance/scripts/set_emm_ip_ranges.py "10.73.79.22, 10.73.79.38, 10.73.71.6"`

### DIFF
--- a/opengever/maintenance/scripts/set_emm_ip_ranges.py
+++ b/opengever/maintenance/scripts/set_emm_ip_ranges.py
@@ -1,0 +1,54 @@
+from opengever.base.ip_range import valid_ip_range
+from opengever.maintenance.debughelpers import setup_app
+from opengever.maintenance.debughelpers import setup_option_parser
+from opengever.maintenance.debughelpers import setup_plone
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/set_emm_ip_ranges.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def set_emm_ip_ranges(ip_ranges):
+    """Search all MS OneNote documents and fixes the contentType.
+    """
+    try:
+        valid_ip_range(ip_ranges)
+        registry = getUtility(IRegistry)
+        settings = registry.forInterface(IOfficeConnectorSettings)
+        settings.office_connector_disallowed_ip_ranges = ip_ranges
```
deiferni, line 20: you could set it easier with https://docs.plone.org/develop/plone.api/docs/portal.html#set-plone-app-registry-record

PR #213, opened 2020-07-09
Defer syncing tasks while renaming objects.

deiferni:
In https://4teamwork.atlassian.net/browse/GEVER-177 we have encountered errors while renaming as task syncing tried to resolve/traverse to objects where the int-id utility did not yet have a chance to update the path stored there, but the object itself had already been renamed.

This lead to a a `KeyError` as described in https://4teamwork.atlassian.net/browse/GEVER-177?focusedCommentId=14815.

With this PR we patch the event handler to make sure that syncing tasks is deferred until all objects have been moved.

### DIFF
--- a/opengever/maintenance/scripts/update_object_ids.py
+++ b/opengever/maintenance/scripts/update_object_ids.py
@@ -18,6 +18,7 @@ Notes:
 """
 
 from opengever.base.interfaces import IReferenceNumber
+from opengever.globalindex.handlers import task as task_handlers
 from opengever.journal import handlers as journal_handlers
 from opengever.maintenance.debughelpers import setup_app
 from opengever.maintenance.debughelpers import setup_plone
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/update_object_ids.py
```diff
@@ -31,18 +32,38 @@
 import transaction
 
 
-class DisabledJournaling(object):
-    """Context manager that temporarily disables creation of journal entries.
+class DeferredOrDisabledEventHandlers(object):
+    """Context manager that temporarily disables or defers or disables events.
```
mbaechtold, line 36: This sentence disables twice ;-)

deiferni:
@mbaechtold i have used the power of small well named methods to hopefully improve readability

mbaechtold:
Nice 👍🏻

PR #218, opened 2020-07-30
Add simple bin/instance run script to update a specific document_type.

phgross:
Script added and checked in mainly for traceability reasons.

Used for https://4teamwork.atlassian.net/browse/GEVER-791.


njohner:
I'm in a merge frenzy...

PR #225, opened 2020-10-12
Add script to reindex path_depth in Solr.

lukasgraf:
Add script to reindex `path_depth` in Solr.

Already executed on PROD, checking it in for traceability.

Jira: https://4teamwork.atlassian.net/browse/CA-957

PR #226, opened 2020-10-16
Add script to reindex is_subtask in solr.

njohner:
It seems that we've had issues during the update from solr 7 to solr 8 for Köniz. This led to fields that were introduced / updated in solr during that update to end up not being updated. It seems that two fields were concerned, `path_depth` and `is_subtask`. `path_depth` was already fixed in https://github.com/4teamwork/opengever.maintenance/pull/225, here we add a similar script to index `is_subtask` where necessary. 

The corresponding upgrade step was https://github.com/4teamwork/opengever.core/blob/master/opengever/core/upgrades/20190827180126_replace_is_subtask_field_index_with_a_boolean_index/upgrade.py

I also update `bootstrap.py` and fix the batching in the script introduced for `path_depth`

For https://4teamwork.atlassian.net/browse/CA-1112

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -25,7 +25,10 @@ import tempfile
 
 from optparse import OptionParser
 
-tmpeggs = tempfile.mkdtemp()
+__version__ = '2015-07-01'
+# See zc.buildout's changelog if this version is up to date.
+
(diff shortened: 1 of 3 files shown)


lukasgraf (approved):
LGTM 👍 , apart from the low hard limit ;)

PR #229, opened 2020-10-26
Add script to check if solr indices in catalog.

deiferni:
We don't seem to have removed them consistently for all deployments. I'd like to quickly check SaaS to get a damage assessment.

jira: https://4teamwork.atlassian.net/browse/CA-1141


njohner (approved):
LGTM

PR #233, opened 2020-11-30
Migrate external reference for workspaces

tinagerber:
Newly, dossiers that are linked to a workspace should be marked with an `ILinkedToWorkspace` interface. In addition, the `external_reference` field of the workspace should contain the `Oguid` of the linked dossier.

Since GEVER and Teamraum are two separate installations, the modifications cannot be migrated automatically. Therefore two migration scripts were written for this purpose.

The first script marks all linked dossiers with the interface. This is done in this script, since all dossiers have to be checked anyway to see if they are linked. Therefore it does not make sense to do this separately in an upgrade step. Also the links dossier <-> workspace are saved in a json file.

The second script reads the generated JSON file and saves the `Oguid` of the linked dossier as `external_reference` of the workspace.

On the GEVER intallation run `bin/instance run ./scripts/mark_dossiers_linked_with_workspaces_and_dump_links.py --dump-directory path/to/directory`

On the teamraum installation run `bin/instance run ./scripts/set_external_reference_for_workspaces.py path/to/file`

Relates to https://github.com/4teamwork/opengever.core/pull/6752

Jira: https://4teamwork.atlassian.net/browse/NE-100

### DIFF
--- a/opengever/maintenance/scripts/set_external_reference_for_workspaces.py
+++ b/opengever/maintenance/scripts/set_external_reference_for_workspaces.py
@@ -1,0 +1,49 @@
+"""
+bin/instance run ./scripts/set_external_reference_for_workspaces.py path/to/file
+Set linked dossier oguid as external_reference for linked workspaces.
+"""
+
+from opengever.maintenance.debughelpers import setup_app
+from opengever.maintenance.debughelpers import setup_option_parser
+from opengever.maintenance.debughelpers import setup_plone
+from plone import api
+import json
+import sys
+import transaction
+
+
+def set_external_reference(file_path):
+    with open(file_path) as file_:
+        data = json.load(file_)
+    uids = data.keys()
+    query = {"object_provides": "opengever.workspace.interfaces.IWorkspace"}
+    catalog = api.portal.get_tool('portal_catalog')
+    workspace_brains = catalog.unrestrictedSearchResults(query)
+    for brain in workspace_brains:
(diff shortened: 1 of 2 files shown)


opengever/maintenance/scripts/set_external_reference_for_workspaces.py
```diff
@@ -0,0 +1,52 @@
...
+    uids = data.keys()
+    query = {"object_provides": "opengever.workspace.interfaces.IWorkspace"}
+    catalog = api.portal.get_tool('portal_catalog')
+    workspace_brains = catalog.unrestrictedSearchResults(query)
+    for brain in workspace_brains:
+        workspace = brain.getObject()
+        if workspace.UID() in uids:
+            workspace.external_reference = data[workspace.UID()]
+        else:
+            workspace.external_reference = ''
```
lukasgraf, line 30: This should be unicode (so, `u''`). A `zope.schema.TextField` always expects unicode, and depending on how it's being set, will complain. Worse, mixed types of unicode vs. bytestrings across different objects can lead to hard to debug errors once they are getting compared to each other or concatenated.

In the PR in opengever.core this was not an issue because everything passes through JSON + the REST API, which already leads to a conversion to unicode.

lukasgraf (changes requested):
LGTM 👍 
Just one minor point.

lukasgraf (approved):

PR #235, opened 2020-12-03
Add tests against 2020.14.5.

deiferni:
🤨


njohner (approved):
huh sorry seems I forgot. (actually I had a look but did not go further than the name of the file, from which I assumed it tested 2020.14-stable. I know in hindsight that is a pretty dumb assumption as we don't always create a stable branch and that tests would not run again when a new version is released...)

PR #237, opened 2020-12-08
Add script to cleanup notifications for ToDo activities.

njohner:
The added script deletes notifications for activities on `ToDo`s that were incorrectly sent. These were due to a bug adding new participant of a `Workspace` as `Watcher` on all `ToDo`s and not just the ones in that particular `Workspace`.

Note that it is not possible to know which `Notification`s were wrongfully sent, as we cannot know what users were legitimate `Watcher`s at the time of the `Activity`. Instead, we delete all notifications for `ToDo` activities that were sent to users who are not currently `Watcher` on the corresponding `Resource`.

For https://4teamwork.atlassian.net/browse/CA-1324

### DIFF
--- a/opengever/maintenance/scripts/cleanup_todo_notifications.py
+++ b/opengever/maintenance/scripts/cleanup_todo_notifications.py
@@ -1,0 +1,85 @@
+"""
+This script deletes notifications for activities on ToDos that were incorrectly
+sent. These were due to a bug adding new participant of a workspace as Watcher
+on all ToDos and not just the ones in that particular workspace.
+
+Note that it is not possible to know which notifications were wrongfully sent,
+as we cannot know what users were legitimate watchers at the time of the
+Activity. Instead, we delete all notifications for ToDo activities that were
+sent to users who are not currently watchers on the corresponding resource.
+
+    bin/instance run ./scripts/cleanup_todo_notifications.py
+
+"""
+from opengever.activity.model import Activity
+from opengever.activity.model import Notification
+from opengever.activity.model import Resource
+from opengever.base.model import create_session
+from opengever.maintenance.debughelpers import setup_app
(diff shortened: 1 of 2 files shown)


opengever/maintenance/scripts/cleanup_todo_notifications.py
```diff
@@ -0,0 +1,82 @@
...
+from opengever.maintenance.debughelpers import setup_option_parser
+from opengever.maintenance.debughelpers import setup_plone
+import sys
+import transaction
+
+
+def find_notifications_to_delete():
+    to_delete = []
+    todo_activities = Activity.query.filter(
+        Activity.kind.in_(['todo-modified', 'todo-assigned'])
```
deiferni, line 27: this is not scoped by the current plone site. will that be an issue? Or is global cleanup desired?

opengever/maintenance/scripts/cleanup_todo_notifications.py
```diff
@@ -0,0 +1,82 @@
...
+        transaction.doom()
+
+    app = setup_app()
+    setup_plone(app, options)
+
+    to_delete = find_notifications_to_delete()
+
+    affected_resources = {notification.activity.resource
+                          for notification in to_delete}
+    affected_objects = {resource.oguid.resolve_object()
```
deiferni, line 67: ah this will be an issue because you are resolving objects here. i'd recommend to limit the query above to resources in the current site.

deiferni (approved):
LGTM 👍  apart from comments.

opengever/maintenance/scripts/cleanup_todo_notifications.py
njohner: thanks. I did want to scope that... I'll change it (although the script is for TR deployments which have only one admin unit ;-))

PR #242, opened 2021-01-18
HBA migration: make sure that all repofolders have an entry in the migration excel.

njohner:
This implements one additional check. Because only entries that are in the excel are actually considered during some of the validations, it is important that all existing repofolders have an entry in the migration Excel. This is much easier than having to check both for positions in the excel and already existing positions (together with moves, merges, etc.).

https://4teamwork.atlassian.net/browse/CA-1266


deiferni (approved):
LGTM 👍

PR #244, opened 2021-01-25
Add script to print trashed document templates.

deiferni:
According to https://4teamwork.atlassian.net/wiki/spaces/4TEAM/pages/123077675/GEVER+Releases+Versions-Spezifische+Punkte we should double-check trashed document templates. We'll use the script suggested in https://github.com/4teamwork/opengever.core/pull/6793#issuecomment-757842941 but check it in so we can run it easily for all deployments.

Jira: https://4teamwork.atlassian.net/browse/CA-1298


lukasgraf (approved):
LGTM 👍

PR #245, opened 2021-01-27
Add script to initialize title_en.

njohner:
We add a new script that initializes the `title_en` field for all concerned portal_types. The script can build up a glossary from a repository excel file and use it to translate the titles of repository roots and folders. For other portal_types it uses a default glossary. When it does not find a given title in the glossary it simply uses the German title instead.

For https://4teamwork.atlassian.net/browse/CA-1040


lukasgraf (approved):
LGTM 👍

PR #246, opened 2021-01-28
Add option to delete trashed document templates.

deiferni:
Add a `--delete` option which allows to automatically remove all trashed templates. As discussed in https://4teamwork.atlassian.net/browse/CA-1298?focusedCommentId=21220  that's the strategy we want to use, at least for SaaS.

Jira: https://4teamwork.atlassian.net/browse/CA-1298


njohner (approved):
LGTM

PR #247, opened 2021-01-28
Improve repository migration scripts

njohner:
We add the possibility to split the task syncing to a different step, as the migration itself is done on an OGDS that is not the productive one. For that we pickle the migrator object which knows which tasks will then have to be synced. This last step will be done in debug mode.

For https://4teamwork.atlassian.net/browse/CA-1267

### DIFF
--- a/opengever/maintenance/scripts/repository_migration.py
+++ b/opengever/maintenance/scripts/repository_migration.py
@@ -296,7 +296,6 @@ class ExcelDataExtractor(object):
         self.data = sheets[0]['sheet_data']
         self.n_data = len(self.data) - self.first_data_row
         self.validate_format()
-        self.is_valid = True
 
     def validate_format(self):
         headers = self.data[self.header_row]
@@ -329,6 +328,7 @@ class RepositoryExcelAnalyser(object):
         self._reference_repository_mapping = None
         self.final_positions = []
         self.catalog = api.portal.get_tool('portal_catalog')
+        self.is_valid = True
 
         # A mapping new_position_number:UID
         self.position_uid_mapping = {}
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/repository_migration.py
```diff
@@ -1179,16 +1246,27 @@ def main():
...
+    analyser.export_to_excel()
+
+    analyser_path = os.path.join(options.output_directory, "analyser.pkl")
+    with open(analyser_path, "w") as outfile:
+        # some attributes cannot be pickled so we simply set them to None
+        analyser.catalog = None
+        analyser.formatter = None
+        analyser.reporoot = None
+        analyser._reference_repository_mapping = None
+        pickle.dump(analyser, outfile)
```
deiferni, line 1269: So, the only thing that you are interested here is the list of UIDs, right? I would advise against pickling the object. IMO you don't gain anything by the pickle, on the contrary it seems to be more complicated at first glance. Wouldn't something simpler like a text file with one item per line be sufficient here? The text file is human readable/editable and does not have any (ok, i admit, theoretical here) potential security impacts.

opengever/maintenance/scripts/repository_migration.py
```diff
@@ -1,3 +1,28 @@
...
+       syncing are stored in the migrator object that is then pickled.
+  -n : dry-run.
+
+If task syncing was skipped, it can be later performed in debug mode:
+
+from opengever.maintenance import dm; dm()
+from opengever.maintenance.scripts.repository_migration import *
+import pickle
+import transaction
+with open(path/to/migrator.pkl, "r") as migrtor_file:
```
deiferni, line 20: migr**a**tor?

deiferni (approved):
LGTM apart from the pickling. Maybe i did not spot all the use-cases here. ☎️  if you wish 😉 .

opengever/maintenance/scripts/repository_migration.py
njohner: Ok I now write some data that might be of interest to json files, as well of course as the list of task UIDs that we will need for syncing...

PR #250, opened 2021-02-17
Add script to initialize title_fr.

njohner:
Very similar to the script initialising english titles. This one is a bit more careful, as French has been around for a while, it takes care not to overwrite French titles that would already have been set.


lukasgraf (approved):
LGTM 👍 

(Did a quick visual diff with [`initialize_title_en.py`](https://github.com/4teamwork/opengever.maintenance/blob/master/opengever/maintenance/scripts/initialize_title_en.py), looks good.

PR #251, opened 2021-02-19
Improve OS migration script

njohner:
- We allow setting new permissions on existing repository folders. Note that we only do this if inheritance is blocked and local_roles are indicated in the Excel. These will replace the existing sharing permissions
- We skip the docproperties update
- We add commits during the migration, this is a test to see whether it will make the migration faster.
- I also improved logging a bit.

Tests were also updated, see https://github.com/4teamwork/opengever.core/pull/6792

For https://4teamwork.atlassian.net/browse/CA-1472

### DIFF
--- a/opengever/maintenance/scripts/repository_migration.py
+++ b/opengever/maintenance/scripts/repository_migration.py
@@ -896,7 +896,7 @@ class RepositoryExcelAnalyser(object):
                 'x' if not data['is_valid'] else '',
                 'x' if data['permissions_disregarded'] else '',
                 'x' if data['local_roles_deleted'] else '',
-                json.dumps(data['permissions']),
+                json.dumps(data['permissions']) if any(data['permissions'].values()) else '',
             ]
 
             for column, attr in enumerate(values, 1):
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/repository_migration.py
```diff
@@ -1013,6 +1013,7 @@ def create_repository_folders(self, items):
         self.start_bundle_import(tmpdirname)
 
         shutil.rmtree(tmpdirname)
+        transaction.commit()
```
lukasgraf, line 1016: These should respect `options.dryrun`.

lukasgraf (changes requested):
LGTM 👍 , except the `commit()` not respecting the dryrun flag.

opengever/maintenance/scripts/repository_migration.py
njohner: True, thanks!

njohner:
All done. As discussed, the last commit will be tested in the next test-migration, see if it really speeds the migration up.

lukasgraf (approved):
LGTM 👍

njohner:
@lukasgraf can you have a look at the last commit? I had to exclude reindexing the SearchableText as we won't have the blobs during migration.

lukasgraf (approved):
LGTM 👍

PR #258, opened 2021-04-06
Also reindex sortable_reference during OS migration.

njohner:
SG is now on the `opengever.core` version 2021.5.1, so we also need to reindex the newly introduced `sortable_reference` index in solr during the OS migration.

For https://4teamwork.atlassian.net/browse/CA-1959


phgross (approved):
👍 makes sense

PR #261, opened 2021-04-28
Don't tests versions no longer used in production.

deiferni:
Drop testing against versions we no longer use in production according to matrix.


njohner (approved):
LGTM

PR #263, opened 2021-05-04
Also use setuptools 44.1.1 for older versions.

deiferni:
We cannot buildout reliably with the old pinned setuptools versions, so we pin the test builds to use a newer version.

I'm expecting most 2019.X versions to go away soon in any case.

This makes sure tests against `master` are green again.


buchi (approved):
👍🏻

PR #265, opened 2021-06-09
Add script to check default values have been persisted on all objects.

njohner:
This is based on the `find_non_persisted_values.py` script.

### DIFF
--- a/opengever/maintenance/scripts/find_fields_with_non_persisted_default_values.py
+++ b/opengever/maintenance/scripts/find_fields_with_non_persisted_default_values.py
@@ -1,0 +1,200 @@
+"""
+Script to find field values that haven't been persisted on objects.
+
+    bin/instance run find_fields_with_non_persisted_default_values.py
+
+This script logs a detailed CSV report and a summary to var/log/, and displays
+some progress info and stats on STDERR/STDOUT.
+"""
+
+from App.config import getConfiguration
+from collections import Counter
+from datetime import datetime
+from opengever.base.default_values import get_persisted_value_for_field
+from opengever.maintenance.debughelpers import setup_app
+from opengever.maintenance.debughelpers import setup_option_parser
+from opengever.maintenance.debughelpers import setup_plone
+from plone import api
+from plone.dexterity.utils import iterSchemataForType
+from zope.component import getUtility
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/find_fields_with_non_persisted_default_values.py
```diff
@@ -0,0 +1,200 @@
...
+        self.stats['by_field'] = Counter()
+
+        ts = datetime.today().strftime('%Y-%m-%d_%H-%M-%S')
+        self.csv_log_path = self.get_logfile_path(
+            'find-nonpersistent-values-%s.csv' % ts)
+        self.summary_log_path = self.get_logfile_path(
+            'find-nonpersistent-values-summary-%s.log' % ts)
+
+    def run(self):
+        sys.stderr.write("Checking for non-persisted values...\n\n")
```
deiferni, line 51: `stderr` does not feel correct here

opengever/maintenance/scripts/find_fields_with_non_persisted_default_values.py
```diff
@@ -0,0 +1,200 @@
...
+                for i, brain in enumerate(all_brains):
+                    obj = brain.getObject()
+                    missing_fields = self.check_for_missing_fields(obj)
+                    self.update_stats(missing_fields)
+
+                    if missing_fields:
+                        self.write_csv_row(obj, missing_fields)
+
+                    if i % 100 == 0:
+                        sys.stderr.write("Progress: %s of %s objects\n" % (i, total))
```
deiferni, line 69: i would prefer a `logger` here, there are examples in other scripts how to set one up that is working. directly writing to stdout/stderr is a bit of an antipattern IMO.

deiferni (approved):
Nothing wrong with the logic IMO, only some nagging about the logging.

opengever/maintenance/scripts/find_fields_with_non_persisted_default_values.py
njohner: Right. I copied the existing script and made minimal adjustments to the logic to only get non persisted fields that actually have a default value. But you're right I'll change that

PR #272, opened 2021-09-01
Add script to set `preserved_as_paper` to False.

deiferni:
Add a small script to set `preserved_as_paper` to False for all `IDocumentMetadata`. We probably could have run it in a debug shell but then again having progress logging and a dry-run mode is useful IMO.

Jira [CA-2772](https://4teamwork.atlassian.net/browse/CA-2772)

### DIFF
--- a/opengever/maintenance/scripts/set_preserved_as_paper_to_false.py
+++ b/opengever/maintenance/scripts/set_preserved_as_paper_to_false.py
@@ -1,0 +1,58 @@
+"""
+Script to set `preserved_as_paper` to False for all `IDocumentMetadata`.
+
+    bin/instance run set_preserved_as_paper_to_false.py
+
+"""
+from opengever.document.behaviors.metadata import IDocumentMetadata
+from opengever.maintenance.debughelpers import setup_app
+from opengever.maintenance.debughelpers import setup_option_parser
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/set_preserved_as_paper_to_false.py
```diff
@@ -0,0 +1,58 @@
...
+        "object_provides": IDocumentMetadata.__identifier__,
+    }
+    brains = catalog.unrestrictedSearchResults(query)
+    total = len(brains)
+
+    for i, brain in enumerate(brains):
+        obj = brain.getObject()
+        meta = IDocumentMetadata(obj)
+        meta.preserved_as_paper = False
+
```
phgross, line 35: I'm not sure, but is there no catalog or solr index?

phgross (approved):
LGTM, 

i was just wondering if a reindexObject is necessary 🤔

opengever/maintenance/scripts/set_preserved_as_paper_to_false.py
deiferni: I was also surprised, but no. There is no index, no catalog metadata and no Solr field.

PR #276, opened 2021-09-15
Add extended health checks (Nightly Jobs and OGDS sync).

lukasgraf:
This uses the `nightly_run_within_24h()` and `ogds_sync_within_24h()` functions introduced in 4teamwork/opengever.core#7176 to provide an extended mode for the `@@health-check` view:

```
GET http://localhost:8080/fd/@@health-check?extended=1

{
   "status":"healthy",
   "instance":{
      "instance_status":"healthy"
   },
   "nightly_jobs":{
      "nightly_jobs_status":"healthy",
      "last_nightly_run":"2021-09-20T17:48:54.225458",
      "job_counts":{
         "maintenance-jobs":5,
         "deliver-sip-packages":0,
         "trigger-missing-archival-file-conversion":0,
         "complete-role-assignment-reports":0,
         "execute-after-resolve-jobs":0
      }
   },
   "ogds":{
      "ogds_sync_status":"healthy",
      "last_ogds_sync":"2021-09-20T17:46:38.552990"
   }
}
```
_(Whitespaced added for clarity - actual response contains all the JSON on one line)_

If any of these two are not ok, their marker strings will be `unhealthy`.

`"instance_status":"healthy"` has the same meaning as `"status": "OK" does in the non-extended version: The instance is alive, and the SQL connection works.

`status` is now an overall-status - it is only `healthy` if all three of Instance, OGDS, and Nightly Jobs are healthy. Otherwise it will be `unhealthy`.

If those helper functions can't be imported from `opengever.core` (because of an old version), they get replaced with stub functions that return `True`. This way we don't get hammered with false-positive service failures. Instead, those checks simply won't be effective yet until `opengever.core` is updated.

The jobs counts are cached (5m) because they do involve quite a few adapter lookups and BTree length checks. On my local machine, both the basic and extended `@@health-checks` took 8ms.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,8 @@ Changelog
 1.0 (unreleased)
 ----------------
 
+- Add extended health checks (Nightly Jobs and OGDS sync). [lgraf]
+
 - Add script to cleanup notifications for ToDo activities. [njohner]
 
 - Add option to index bumblebee checksums. [buchi]
--- a/opengever/maintenance/browser/health_check.py
+++ b/opengever/maintenance/browser/health_check.py
@@ -1,4 +1,6 @@
+from plone.memoize import ram
 from Products.Five.browser import BrowserView
+from time import time
 import json
 
 try:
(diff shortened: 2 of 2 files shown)


njohner (approved):
LGTM

PR #278, opened 2021-09-22
Add script to help creating an ldapsync config file.

njohner:
To be able to configure the AppSwitcher correctly in Teamraum deployments and display all available AdminUnits/OrgUnits, we need to sync all the corresponding groups to the Teamraum LDAP. Writing the configuration file for the LDAP sync can be a pain for deployment with man such OrgUnits. This script will simply print out that part of the ldapsync config file, i.e. source and target groups for all OrgUnits.

For https://4teamwork.atlassian.net/browse/CA-2821


phgross (approved):
looks good

PR #285, opened 2021-11-15
Add script to add InterGEVER webaction.

lukasgraf:
This script adds a webaction on dossiers for eCH-0147 export via InterGEVER.

If a webaction with the same title already exists, the script skips the registration to avoid duplicate actions.

For [CA-2789](https://4teamwork.atlassian.net/browse/CA-2789)


deiferni (approved):
You could also use `unique_name` to control only creating once.

lukasgraf:
@deiferni good point, thanks! Updated accordingly.

PR #290, opened 2021-12-03
Add maintenance script to propagate restricted field values to children.

njohner:
Add maintenance script updates the values of the passed restricted field for all children of a given container, to make sure that the values for that field satisfy the restrictions.

https://4teamwork.atlassian.net/browse/CA-3239

### DIFF
--- a/opengever/maintenance/scripts/propagate_restricted_field_value.py
+++ b/opengever/maintenance/scripts/propagate_restricted_field_value.py
@@ -1,0 +1,161 @@
+"""
+The script updates the values of the passed restricted field for all children
+of a given container, to make sure that the values for that field satisfy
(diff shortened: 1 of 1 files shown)


phgross (approved):
Scripts LGTM, but I'm not sure if fulfill the customer request. I have the feeling, that they expect to change all objects no matter if the old value would be still valid or not. 

But i am not sure here either and maybe it doesn't matter anyway because all the classification data is invalid.

njohner:
@phgross Good point I'll ask the customer what he expects

njohner:
@phgross Ok I asked and they want to keep valid values as is (i.e. more restrictive values than the new one remain as they are). So I'll merge.

PR #294, opened 2022-01-25
Add script to set an arbitrary registry record to a given value.

lukasgraf:
Add script to set an arbitrary registry record to a given value.

Expects the command line argument `<value>` to be given as a string representing a Python literal, which is then parsed using `literal_eval`. 

Example:

```
bin/instance run set_registry_value.py opengever.base.interfaces.IGeverUI.is_feature_enabled True
```

For [CA-3456](https://4teamwork.atlassian.net/browse/CA-3456)

### DIFF
--- a/opengever/maintenance/scripts/set_registry_value.py
+++ b/opengever/maintenance/scripts/set_registry_value.py
@@ -1,0 +1,67 @@
+"""
+Allows to set an arbitrary registry record to a particular value.
+
+The <value> is supposed to be specified as a Python literal.
+
+Example Usage:
+
+    bin/instance run set_registry_value.py <record_dotted_name> <value>
+    bin/instance run set_registry_value.py opengever.base.interfaces.IGeverUI.is_feature_enabled True
+"""
+from ast import literal_eval
+from opengever.maintenance.debughelpers import setup_app
+from opengever.maintenance.debughelpers import setup_plone
+from plone.registry.interfaces import IRegistry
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/set_registry_value.py
```diff
@@ -0,0 +1,67 @@
...
+import argparse
+import sys
+import transaction
+
+
+def cast_value(value, pytype):
+    value = literal_eval(value)
+    if isinstance(value, pytype) or value is None:
+        return value
+
```
njohner, line 25: didn't you want to try to cast here: `pytype(value)`?

njohner (approved):
LGTM

opengever/maintenance/scripts/set_registry_value.py
lukasgraf: Something along those lines is what I first went with, but:
- `bool(4)` would happily cast to `True`
- plone.registry happily accepts a value of `4` for a `bool` record too
- As would `bool("False")`
- Casting wouldn't work at all for something like `[1, 2, 3]`, but `literal_eval` can handle that just fine

So, given all these, I went with an approach that first parses the Python literal (or rejects it, if not a valid literal), and then asserts that it is exactly of the expected type (which it should be without any further casting).

This makes it more strict, and avoids ambigous cases like `int(True)` or `bool(1)` being casted implicitly, which probably mean someone is using it wrong.

PR #296, opened 2022-01-27
Improve repository migration script

njohner:
To finally get a few benefits, I had to modify the migration script quite a bit. Before we were generating a mapping of old and new positions in the order of the rows in the excel file and at the same time analysing the corresponding operation. This made the whole script very sensitive to operation ordering in the excel file, and defining the correct order was really not always easy (as the operations themselves actually happen in the order `create`, `move`, `merge`). Moreover we were sometimes using `position`s (i.e. reference numbers) to identify the parents (for example for creation operations). All the creation operations happening first,  can lead to temporarily having 2 repository folders with the same reference number (e.g. are creating a new repofolder having the same number as an already existing one which will be moved later in the migration), which in turn would lead to issues identifying the correct parent in other creation operations. To solve these issues (and also better identify some move operations), we now:
- assign a `guid` to all existing repository folders
- construct two mappings:
  - old (initial) positions to its guid
  - new (final) positions to their guid (we assign a guid here to positions that will be created later)
- Once this is done we go a second time over all the rows in the excel and identify to what operation it will correspond

With this it is easier to correctly identify different operation types and the identification is less dependent on the order in the excel file.

Also note that I've updated the corresponding tests in opengever.core, to make sure the migration script was still working as before and added a new test for the newly supported edge cases (see https://github.com/4teamwork/opengever.core/pull/6792, last 5 commits)

For https://4teamwork.atlassian.net/browse/CA-3245 and https://4teamwork.atlassian.net/browse/CA-3243

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,8 @@ Changelog
 1.0 (unreleased)
 ----------------
 
+- Handle new argument "graceful" of task syncer in ObjectIDUpdater. [njohner]
+
 - Add extended health checks (Nightly Jobs and OGDS sync). [lgraf]
 
 - Add script to cleanup notifications for ToDo activities. [njohner]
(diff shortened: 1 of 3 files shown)


elioschmutz (approved):
👍

PR #299, opened 2022-02-08
Add script to check workspace folder participation validity.

njohner:
This script allows to verify that local roles on the migrated objects satisfy the restrictions we have in Teamraum, i.e. only users with permissions on a parent object can have local roles on a child object, and only objects with blocked inheritance can have local roles.

This is for https://4teamwork.atlassian.net/browse/CA-1481

### DIFF
--- a/opengever/maintenance/scripts/check_workspace_participants.py
+++ b/opengever/maintenance/scripts/check_workspace_participants.py
@@ -1,0 +1,70 @@
+"""
+Script that checks that participants set on workspace folders are valid,
+i.e. are also participants on their parent.
+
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/check_workspace_participants.py
```diff
@@ -0,0 +1,70 @@
...
+import sys
+
+
+class ParticipantsChecker(object):
+
+    def __init__(self):
+        self.misconfigured = TextTable()
+        self.misconfigured.add_row((
+            u"Pfad",
+            u"Fehlender Zugriff einschr\xe4nkung",
```
lukasgraf, line 23: `Fehlende Zugriffseinschr\xe4nkung` sounds a bit better to my ear.

PR #300, opened 2022-02-09
Set Metadata when creating new positions in repository migration. 

njohner:
We add support for the metadata fields (`IClassification` and `ILifeCycle` behaviours) for repository folders created during repository migration.

For [CA-3245]

[CA-3245]: https://4teamwork.atlassian.net/browse/CA-3245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/maintenance/scripts/repository_migration.py
+++ b/opengever/maintenance/scripts/repository_migration.py
@@ -6,7 +6,7 @@ migration.
 optional arguments:
   -o : path to a folder where output should be saved. The folder will be created.
        This defaults to var/migration-TIMESTAMP
-  -s : sync tasks. By default tasks are not synced as the migration is not
+  -t : sync tasks. By default tasks are not synced as the migration is not
        performed on the original OGDS. Instead UIDs of tasks that will need
        syncing are stored and dumped in a json file that can then be used
        to sync the tasks.
--- a/opengever/maintenance/scripts/repository_migration.py
+++ b/opengever/maintenance/scripts/repository_migration.py
@@ -38,6 +38,7 @@ from Acquisition import aq_parent
 from collections import defaultdict
 from collections import namedtuple
 from collective.transmogrifier.transmogrifier import Transmogrifier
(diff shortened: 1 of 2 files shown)


opengever/maintenance/scripts/repository_migration.py
```diff
@@ -859,6 +894,35 @@ def extract_permissions(self, row):
 
         return permissions
 
+    def extract_metadata(self, row):
+        metadata = {}
+        for field in metadata_fields:
+            fieldname = field.getName()
+            value = getattr(row, fieldname)
```
lukasgraf, line 913: `getattr()` without an explicit default as the third argument will raise `AttributeError` (depending on the object - not exactly sure what type `row` is here). Just making sure this is deliberate.

opengever/maintenance/scripts/repository_migration.py
```diff
@@ -895,6 +959,9 @@ def insert_label_row(self, sheet):
             'Verletzt Leafnode Prinzip',
             'Ist ungultig',
 
+            # metadata
+            'Metadaten Nichtueberstimmung',
```
lukasgraf, line 979: That doesn't quite work in German 😉  I'd say `Metadaten stimmen nicht ueberein`, unless you need a noun-ified message. Then I'd say `Metadaten-Aenderung`.

lukasgraf (approved):
LGTM! 👍 
Just some trivialities.

opengever/maintenance/scripts/repository_migration.py
njohner: row has an attribute for every key in `metadata_fields`.

opengever/maintenance/scripts/repository_migration.py
njohner: 😅

PR #304, opened 2022-02-16
Sort misconfigured objects by path in ParticipantsChecker.

njohner:
Sorting makes the output a bit easier to read. I've now also added a new column with the titles of the objects, that also helps the customer assess what to do with these objects.

For https://4teamwork.atlassian.net/browse/CA-3712


lukasgraf (approved):
LGTM 👍

PR #305, opened 2022-02-21
Implement fixing of participations for migrated teamraum deployments.

njohner:
We modify the script used to check whether local roles respect the restrictions imposed in the new teamraum to allow correcting them. What the script will do:
- Correct participations with multiple roles: each participant should only have a single role (we keep the one giving the most permissions)
- Remove participants with same roles (or lower permissions) as on parent and inheritance that is not blocked.
- Delete participants that are not participants on the parent object
- Block inheritance on objects with local roles and copy roles from parent.

For https://4teamwork.atlassian.net/browse/CA-3712 and https://4teamwork.atlassian.net/browse/CA-3713

### DIFF
--- a/opengever/maintenance/scripts/check_workspace_participants.py
+++ b/opengever/maintenance/scripts/check_workspace_participants.py
@@ -202,7 +202,7 @@ class ParticipantsChecker(object):
                 self.misconfigured.add_row((
                     self.get_url(obj),
                     self.get_titles(obj),
-                    'x' if missing_admin else'',
+                    'x' if missing_admin else '',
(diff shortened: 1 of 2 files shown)


lukasgraf (approved):
LGTM 👍 
But I gotta be honest, I don't think I quite understand the fix for the missing roles block. Maybe we can have a quick 📞 about that one?

njohner:
@lukasgraf I've updated the PR with your suggestions and had to fix a few more things (see the last 4 commits).
Can you have another look?

lukasgraf (approved):
LGTM 👍

PR #313, opened 2022-03-17
Add script to globally assign role to principal.

lukasgraf:
Maintenance script that allows to globally assign a role to a principal (via `portal_role_manager`).

For [CA-2880](https://4teamwork.atlassian.net/browse/CA-2880)


njohner (commented):
LGTM

PR #314, opened 2022-03-18
Add script to set nightly jobs time window.

lukasgraf:
Add script to set nightly jobs time window.

(This is required because the existing `set_registry_value.py` script parses / casts any Python literals - but something like `datetime.timedelta(hours=23)` can't be expressed as a Python literal).

For [CA-2880](https://4teamwork.atlassian.net/browse/CA-2880)


njohner (approved):
LGTM

PR #315, opened 2022-03-28
Extend script to create InterGEVER service user and assign roles

lukasgraf:
This extends the script to connect a GEVER deployment to InterGEVER with
- creation of the `intergever.app` service user (if it doesn't already exist)
- assigning the required roles (if not already assigned yet)

I already tested this script like this on `dev.onegovgever.ch`

For [CA-3274](https://4teamwork.atlassian.net/browse/CA-3274)

### DIFF
--- a/opengever/maintenance/scripts/connect_to_intergever.py
+++ b/opengever/maintenance/scripts/connect_to_intergever.py
@@ -3,7 +3,7 @@ Adds a webaction on Dossiers to launch eCH-0147 export via Intergever.
 
 Example Usage:
 
-    bin/instance run add_intergever_webaction.py sgtest
+    bin/instance run connect_to_intergever.py sgtest
 """
 from opengever.api.validation import get_validation_errors
 from opengever.maintenance.debughelpers import setup_app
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/connect_to_intergever.py
```diff
@@ -0,0 +1,133 @@
...
+    if NOTIFICATION_ROLE not in valid_roles:
+        raise Exception(
+            "Role %r not found! Is your GEVER deployment up to date? "
+            "(At least version 2022.7 is required)" % NOTIFICATION_ROLE
+        )
+
+    service_user = api.user.get(SERVICE_USER_ID)
+    existing_roles = role_manager.getRolesForPrincipal(service_user)
+
+    roles_to_assign = [NOTIFICATION_ROLE]
```
deiferni, line 111: Just to make sure as i'm not certain any more. This is the only role currently required for the user 🤔? The rest is done via impersonation and the "real" users roles?

deiferni (approved):
LGTM 👍

opengever/maintenance/scripts/connect_to_intergever.py
lukasgraf: Exactly. I locally had also given it `Member` and `Editor`, but it turns out those aren't necessary. I therefore followed the "least privileges" approach, and I suggest we add only more roles as they become necessary.

PR #318, opened 2022-04-14
Add script to make an entire repository read-only.

lukasgraf:
This adds a script that makes an entire repository read only by removing roles other than `Reader` from local role assignments.

The script's implementation deviates slightly (for repo roots) from what I specified in the story:

> Auf dem Ordnungssystem-Root sollen die Berechtigungseinstellungen fix wie folgt gesetzt werden:
>
> - Benutzer-Gruppe des Mandanten: Lesen
> - [x] Berechtigungen von übergeordneten Ordnern übernehmen
> 
> Möglicherweise vorhandene andere lokale Rollen sollen entfernt werden.

Instead, it does exactly the same thing for all nodes, repository root or otherwise: Remove all roles except `Reader` from any local role assignments. This gurantees that the `View` permission doesn't change, and we never have to reindex security, making the script much faster. In practice, this should not make a difference for the customer - they want to make the repository read-only, not clean up local role assignments (the suggestion to tread the repository root differently was mine anyway, that wasn't something they asked for).

For [CA-3701](https://4teamwork.atlassian.net/browse/CA-3701)

### DIFF
--- a/opengever/maintenance/scripts/make_repository_readonly.py
+++ b/opengever/maintenance/scripts/make_repository_readonly.py
@@ -46,6 +46,10 @@ def get_path(obj):
     return "/".join(obj.getPhysicalPath())
 
 
+def provides_any(obj, iface_list):
+    return any([iface.providedBy(obj) for iface in iface_list])
+
+
 def setup_logger():
     logger = getLogger()
     logger.setLevel(logging.INFO)
@@ -73,6 +77,13 @@ class MakeRepositoryReadOnly(object):
         u"Publisher",
         u"Reviewer",
     )
+
+    OPEN_STATES = DOSSIER_STATES_OPEN + OPEN_TASK_STATES
+
+    CLOSED_STATES = DOSSIER_STATES_CLOSED + CLOSED_TASK_STATES + [
+        'task-state-skipped',
+        'task-state-planned',
+    ]
 
     def __init__(self, plone, args):
         self.plone = plone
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/make_repository_readonly.py
```diff
@@ -0,0 +1,363 @@
...
+            logger.error("The following %s are not closed:" % type_name)
+            for brain in open_objects:
+                logger.error(brain.getPath())
+            raise Exception("Found open %s!" % type_name)
+
+    def assert_in_closed_state(self, obj):
+        """Check again that dossiers and tasks are closed (using the obj).
+        """
+        if IDossierMarker.providedBy(obj):
+            if api.content.get_state(obj) not in DOSSIER_STATES_CLOSED:
```
njohner, line 154: dossier-state-offered is neither in the `DOSSIER_STATES_CLOSED` nor in the `DOSSIER_STATES_OPEN`

opengever/maintenance/scripts/make_repository_readonly.py
```diff
@@ -0,0 +1,363 @@
...
+                raise Exception("Unexpected open dossier: %r" % obj)
+
+        if ITask.providedBy(obj):
+            if api.content.get_state(obj) not in CLOSED_TASK_STATES:
```
njohner, line 158: `task-state-planned` and `task-state-skipped` are neither in `CLOSED_TASK_STATES` not in `OPEN_TASK_STATES`

opengever/maintenance/scripts/make_repository_readonly.py
```diff
@@ -0,0 +1,363 @@
...
+
+    More precisely it makes sure that for every object where we (potentially)
+    modified the local roles, the list of principals that have 'View' is the
+    same before and after the change.
+
+    This is an assumption that we make that allows us to skip reindexing of
+    allowedRolesAndUsers, therefore making the script much faster.
+
+    It's also a business level constraint for this script: Just "making a
+    repository only" should not result in any other users having View access
```
njohner, line 273: "making a repository only" -> "making a repository read-only"

njohner (commented):
I found one issue that I think you need to correct, namely not all states are in the `open` or `closed` sets, which will lead to failures...
The rest is a few small comments

lukasgraf:
@njohner thanks for the thorough review!
I added handling for the task states "planned" and "skipped", and am considering them "closed" for the purpose of this script:

- Tasks can't be edited or added to in those workflow states
- "planned" can only exist inside an open task anyway, which would already block the script

The "dossier-state-offered" I didn't handle explicitly, because it shouldn't be relevant for ZG, but made sure unexpected (=unmapped) review states won't go unnoticed in the future (see commit message of last commit for more details).

Would you mind having another quick look (just the last commit)?

opengever/maintenance/scripts/make_repository_readonly.py
```diff
@@ -146,13 +157,19 @@ def _assert_no_open_objects(self, type_name, type_iface, open_states):
...
 
-        if ITask.providedBy(obj):
-            if api.content.get_state(obj) not in CLOSED_TASK_STATES:
-                raise Exception("Unexpected open task: %r" % obj)
+        review_state = api.content.get_state(obj)
+
+        if review_state not in self.CLOSED_STATES + self.OPEN_STATES:
+            raise Exception(
+                "Unexpected review state %r for object %r! Please handle this "
+                "review state it by making sure it's listed in either "
```
njohner, line 168: review state it by making sure -> review state by making sure

njohner (approved):
LGTM

PR #323, opened 2022-06-09
Add support for modifying or removing local roles.

njohner:
We do no longer only set local roles but also allow removing them. Moreover we fix a bug that led to only setting local roles when inheritance was blocked.
I've updated the existing tests and added a new one, see https://github.com/4teamwork/opengever.core/commits/nj/CA-1266/os_migration

For https://4teamwork.atlassian.net/browse/CA-4293


opengever/maintenance/scripts/repository_migration.py
```diff
@@ -765,18 +764,34 @@ def validate_operation(self, operation):
...
 
+    def are_permissions_updated(self, obj, permissions, inheritance_blocked):
+        if inheritance_blocked != getattr(obj, '__ac_local_roles_block__', False):
+            return True
+        old_permissions = {role_shortname: [] for role_shortname in managed_roles_shortnames}
+        assignments = RoleAssignmentManager(obj).get_assignments_by_cause(ASSIGNMENT_VIA_SHARING)
+        for assignment in assignments:
+            for role in assignment['roles']:
+                shortname = SHORTNAMES_BY_ROLE[role]
+                old_permissions[shortname].append(assignment['principal'])
```
elioschmutz, line 782: `shortname` can be any role from the assignment. But the `old_permissions` dict only provides `managed_roles_shortnames` which is a subset of `SHORTNAMES_BY_ROLE`. I think this will not work

opengever/maintenance/scripts/repository_migration.py
```diff
@@ -765,18 +764,34 @@ def validate_operation(self, operation):
...
+                pass
+            elif self.are_permissions_updated(obj, permissions, inheritance_blocked):
+                operation['set_permissions'] = True
+                if RoleAssignmentManager(obj).get_assignments_by_cause(ASSIGNMENT_VIA_SHARING):
                     operation['local_roles_deleted'] = True
                     logger.warning(
                         "\nSharing assignments for {} will be deleted and "
                         "replaced.\n".format(obj.absolute_url_path()))
 
+    def are_permissions_updated(self, obj, permissions, inheritance_blocked):
```
elioschmutz, line 774: I thought it means if that if the permissions are already updated... what do you think about `needs_permission_update`?

elioschmutz (approved):
Looks good, but you should take a closer look at the `are_permissions_updated` function. I think this will not work in all cases...

opengever/maintenance/scripts/repository_migration.py
```diff
@@ -24,8 +24,6 @@
 transaction.commit()
 
 Notes:
-- Permissions are only taken into account if some local_roles are set (i.e. it
-  is currently not possible to set local_roles back to None)
```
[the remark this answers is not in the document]
njohner: It was not implemented because back then, I did not expect that the migration excel would be complete / up to date with the state of the productive system, so I did not want to overwrite permissions that were added by hand. It turns out, that the excels are up to date and they expect the state after the migration to reflect the data in the migration excel.

njohner:
Hey Elio, I had a but when deleting all local roles on positions, only the sharing assignments would get cleared, but the `__ac_local_roles__` would not get updated. I've fixed this in a new commit: https://github.com/4teamwork/opengever.maintenance/pull/323/commits/3616d536f5b5e76ec0e18479e3e9a4c499ba4d39. Can you have another look?

PR #325, opened 2022-06-22
Fix reference number mappings after os migration

njohner:
There were several issues with the reference number mappings after the OS migrations:
- Wrong reference numbers were used when regenerating the reference number mappings after changing reference numbers. The issue was that we were getting the reference number prefix from the object, when it is actually stored in the annotations, so we need to make sure to access it via the behaviour...
- Some creation operations left behind reference numbers as inactive in the mapping, because the positions were first created with a default number and then the number is set.
- Move operations do not seem to update the mapping on the old parent correctly.

As usual some new tests were added to https://github.com/4teamwork/opengever.core/tree/nj/CA-1266/os_migration

For https://4teamwork.atlassian.net/browse/CA-4356

### DIFF
--- a/opengever/maintenance/scripts/repository_migration.py
+++ b/opengever/maintenance/scripts/repository_migration.py
@@ -1217,15 +1217,11 @@ class RepositoryMigrator(MigratorBase):
         logger.info("\n\nRegenerating number mappings...\n")
         for obj in objs:
             ref_adapter = IReferenceNumberPrefix(obj)
-            # This purges also the dossier mapping, but the parents does not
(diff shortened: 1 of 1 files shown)


phgross (approved):
👍  LGTM.

_So we have actually thought about it, but I implemented it wrong 🙈_

phgross (approved):
LGTM 👍 

_I always use the `ProgressLogger` from ftw.upgrade for progress logging_

njohner:
I couldn't test my last fix on a real migration (it failed before that fix and then I ran the migration without this PR).

PR #328, opened 2022-06-30
Handle duplicate participation entries.

phgross:
In earlier GEVER versions it was apparently possible to store multiple
participations for one contact or org_role. On the production system we have such duplicate participations.

Additional fix for #327 ([CA-2536](https://4teamwork.atlassian.net/browse/CA-2635))


njohner (approved):
Yes that rings a bell...
LGTM

PR #333, opened 2022-08-03
Add simple script to add propertysheets from the console.

phgross:
it expects a json file with the corresponding informations. An example json file is added.

so we can only add the propertysheets.json to the policies and don't need to copy & paste the same script in all policies.


lukasgraf (approved):
LGTM 👍

PR #336, opened 2022-08-19
Nj/ca 3912/migrate spv

njohner:
For https://4teamwork.atlassian.net/browse/CA-3912

### DIFF
--- a/opengever/maintenance/scripts/migrate_and_deactivate_meetings.py
+++ b/opengever/maintenance/scripts/migrate_and_deactivate_meetings.py
@@ -1,0 +1,41 @@
+"""
+This script is used to prepare deletion of the meetings from
+Gever and use RIS in its stead.
+"""
+from opengever.maintenance.debughelpers import setup_app
+from opengever.maintenance.debughelpers import setup_option_parser
+from opengever.maintenance.debughelpers import setup_plone
+from plone import api
+import logging
+import transaction
+
+logger = logging.getLogger('opengever.maintenance')
+logging.root.setLevel(logging.INFO)
+
+
+class MeetingsContentMigrator(object):
+
+    def __init__(self):
+        self.catalog = api.portal.get_tool("portal_catalog")
+
+
+def main():
+    app = setup_app()
+    setup_plone(app)
+
+    parser = setup_option_parser()
+    parser.add_option("-n", dest="dryrun", action="store_true", default=False)
+    (options, args) = parser.parse_args()
+
+    if options.dryrun:
+        logger.info('Performing dryrun!\n')
+        transaction.doom()
+
+    migrator = MeetingsContentMigrator()
+
+    if not options.dryrun:
+        transaction.commit()
+
+
+if __name__ == '__main__':
+    main()
--- a/opengever/maintenance/scripts/migrate_and_deactivate_meetings.py
+++ b/opengever/maintenance/scripts/migrate_and_deactivate_meetings.py
@@ -3,6 +3,9 @@ This script is used to prepare deletion of the meetings from
 Gever and use RIS in its stead.
 """
 from Acquisition import aq_parent
+from collective.taskqueue.interfaces import ITaskQueue
+from collective.taskqueue.interfaces import ITaskQueueLayer
+from collective.taskqueue.taskqueue import LocalVolatileTaskQueue
 from opengever.base.model.favorite import Favorite
 from opengever.base.oguid import Oguid
 from opengever.base.transport import BASEDATA_KEY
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/migrate_and_deactivate_meetings.py
```diff
@@ -106,6 +110,53 @@ def replace_meeting_dossier_with_normal_dossier(self):
...
+                    title=agendaitem.get_title(include_number=True, formatted=True),
+                    container=meeting_dossier)
+
+                # Copy documents from submitted proposal into subdossier
+                for doc in agendaitem.get_documents():
+                    self.copy_and_add_to_mapping(doc, dossier)
+
+                # Move excerpts
+                for doc in agendaitem.get_excerpt_documents():
+                    self.move_and_add_to_mapping(doc, dossier)
```
lukasgraf: Maybe add an `assert doc in meeting_dossier.objectValues()` here, to make sure we're only moving documents from the main (new) meeting dossier into one of its subdossiers, and not from anywhere else? (Which would indicate something unexpected is going on).

opengever/maintenance/scripts/migrate_and_deactivate_meetings.py
```diff
@@ -106,6 +110,53 @@ def replace_meeting_dossier_with_normal_dossier(self):
...
+
+                # Move the proposal document
+                document = agendaitem.resolve_document()
```
lukasgraf, line 203: Same here, could assert that to document to be moved is actually in `meeting_dossier.objectValues()`

opengever/maintenance/scripts/migrate_and_deactivate_meetings.py
```diff
@@ -0,0 +1,303 @@
...
+            migration_annotations = self.get_migration_annotations(dossier)
+            migration_annotations['state'] = api.content.get_state(meeting_dossier)
+
+            # We will delete the dossiers afterwards, as it otherwise seems
+            # to modify meeting_dossiers during iteration.
+            to_delete.append(meeting_dossier)
+
+        for obj in to_delete:
+            # delete meeting_dossier
+            api.content.delete(obj)
```
lukasgraf, line 177: I would probably do some high level logging here, e.g.
`Deleting meeting dossier {meeting_dossier_path} after moving contents to regular dossier {new_dossier_path}`

opengever/maintenance/scripts/migrate_and_deactivate_meetings.py
```diff
@@ -0,0 +1,303 @@
...
+                # Move excerpts
+                for doc in agendaitem.get_excerpt_documents():
+                    self.move_and_add_to_mapping(doc, dossier, meeting_dossier)
+
+                # Move the proposal document
+                document = agendaitem.resolve_document()
+                self.move_and_add_to_mapping(document, dossier, meeting_dossier)
+
+    def set_meeting_dossier_state(self):
+        for meeting in Meeting.query:
```
lukasgraf, line 207: This could also be invoked from `replace_meeting_dossier_with_normal_dossier()` (after setting the state), which would avoid having to find the new "meeting" dossiers by re-querying.

I'm not worried about performance, but the fact that this `Meeting.query` relies on SQLAlchemy returning us the _new_ meeting dossiers, after their reference has been updated, but not been committed yet. In other words, it depends on [dirty reads](https://www.sqlshack.com/dirty-reads-and-the-read-uncommitted-isolation-level/).

I don't think it's a huge deal, but it sort of makes us more reliant on particular isolation levels, and might possibly behave differently on Oracle.

lukasgraf (approved):
LGTM! 👍 
Just some minor comments.

opengever/maintenance/scripts/migrate_and_deactivate_meetings.py
njohner: Right thanks, I had actually planed to add more logging overall and seems I forgot. I'll use the progress logger everywhere and add some explicit logging what is deleted and where stuff gets moved to. I'll also add the path of the former meeting dossier into the annotations of the new dossier and check that the object being deleted is empty...

opengever/maintenance/scripts/migrate_and_deactivate_meetings.py
njohner: I'm not sure that would work, as I would then need to move stuff from the agendaitems into dossiers which might be resolved or inactive, which would lead to permission issues. I'll store the list of dossiers.
So I've reworked the script to iterate over the meetings just once and complete the migration of a single meeting before moving to the next.

opengever/maintenance/scripts/migrate_and_deactivate_meetings.py
njohner: No the proposal document can be either in the submitted proposal or in the meeting dossier (for ad hoc proposals).

lukasgraf (approved):
Sorry, forgot to give you feedback yesterday - the latest commits LGTM! 👍

deiferni (commented):
as just discussed:
- make sure to keep or remove locks consistently (keep excerpts, remove other ones)
- double check why proposal document is moved and not copied
- double check if the meeting flag should be used to prevent creation of meeting content in "Sitzungen"

lukasgraf (approved):
LGTM 👍

PR #339, opened 2022-09-16
Add script to export and close all dossiers inside a repofolder.

njohner:
For https://4teamwork.atlassian.net/browse/CA-4515

### DIFF
--- a/opengever/maintenance/scripts/export_and_close_dossiers.py
+++ b/opengever/maintenance/scripts/export_and_close_dossiers.py
@@ -42,9 +42,10 @@ class DossierExporter(object):
     allowed_final_states = ['dossier-state-resolved', 'dossier-state-inactive']
     states_to_resolve = ['dossier-state-active']
 
-    def __init__(self, context, output_directory, check_only=False):
+    def __init__(self, context, output_directory, check_only=False, dont_close_dossiers=False):
         self.check_only = check_only
         self.output_directory = output_directory
+        self.dont_close_dossiers = dont_close_dossiers
         self.context = context
         self.catalog = api.portal.get_tool("portal_catalog")
 
@@ -60,7 +61,8 @@ class DossierExporter(object):
         if self.check_only:
             return
 
-        self.resolve_dossiers()
+        if not self.dont_close_dossiers:
+            self.resolve_dossiers()
 
         # create output directory
         os.mkdir(self.output_directory)
@@ -68,6 +70,11 @@ class DossierExporter(object):
 
     def check_preconditions(self):
         logger.info("Checking preconditions...")
+
+        if self.dont_close_dossiers:
+            logger.info("dont_close_dossiers: Skip checking dossier state and resolvability")
+            return
+
         # All dossiers should be either inactive, resolved or active and resolvable.
         unresolvable_dossiers = []
         dossiers_in_bad_state = []
@@ -117,7 +124,7 @@ class DossierExporter(object):
     def export_dossiers(self):
         message = "Exporting dossiers."
         for brain in ProgressLogger(message, self.dossiers, logger):
-            if brain.review_state not in self.allowed_final_states:
+            if brain.review_state not in self.allowed_final_states and not self.dont_close_dossiers:
                 raise DisallowedReviewState("Dossier in disallowed review state")
             self._export_dossier(brain.getObject())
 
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/export_and_close_dossiers.py
```diff
@@ -0,0 +1,213 @@
...
+            path=self.context.absolute_url_path(),
+            object_provides=IDossierMarker.__identifier__,
+            is_subdossier=False)
+
+    def __call__(self):
+        self.check_preconditions()
+        if self.check_only:
+            return
+
+        if self.dryrun:
```
lukasgraf, line 64: We usually do this in `main()`, outside the class that does the work. Is there a particular reason this needs to be moved inside? (Having it outside would guard against the resolver issuing a commit as a side effect. But maybe that's exactly the reason it's here, because IIRC the resolve lock gets committed immeditately).

opengever/maintenance/scripts/export_and_close_dossiers.py
```diff
@@ -0,0 +1,213 @@
...
+        folder_path = self._get_output_path(self.output_directory, dossier.title)
+        os.mkdir(folder_path)
+        res = self.catalog.unrestrictedSearchResults(
+            path=dossier.absolute_url_path(),
+            object_provides=IBaseDocument.__identifier__)
+        for brain in res:
+            doc = brain.getObject()
+            filename, ext = os.path.splitext(doc.get_filename())
+            file_path = self._get_output_path(folder_path, filename, ext)
+            with open(file_path, "w") as fout:
```
lukasgraf, line 154: This could be done a bit more efficiently by coping the blob on the file system, instead of reading it into memory and dumping it again.
`doc.file._blob.committed()` should give you the path to the blob in blob storage, and that could then be copied using `shutil.copy2()` (which, incidentally, should also preserve last modification date).

opengever/maintenance/scripts/export_and_close_dossiers.py
```diff
@@ -0,0 +1,213 @@
...
+
+        # create output directory
+        os.mkdir(self.output_directory)
+        self.export_dossiers()
+
+        if not self.dryrun:
+            logger.info("Committing...")
+            transaction.commit()
+
+        logger.info("Done!")
```
lukasgraf, line 78: Log output directory where the export got written to?

opengever/maintenance/scripts/export_and_close_dossiers.py
```diff
@@ -0,0 +1,213 @@
...
+    def __call__(self):
+        self.check_preconditions()
+        if self.check_only:
+            return
+
+        if self.dryrun:
+            logger.info('Performing dryrun!\n')
+            transaction.doom()
+
+        self.resolve_dossiers()
```
lukasgraf, line 68: In order for the script to also be usable for https://4teamwork.atlassian.net/browse/CA-4459 , resolving dossiers (and checking preconditions) should be optional. Could you add a flag `--dont-close-dossiers` that, when given, would skip those two steps?

(I think we may have discussed this in a refinement while you were on vacation. For CA-4459 we decided to not do an estimation, and instead make this script reusable for their case. We linked the stories, but didn't really spell this out in the acceptance criteria 😢 )

lukasgraf (changes requested):
LGTM 👍 , apart from the comments.

opengever/maintenance/scripts/export_and_close_dossiers.py
njohner: No indeed, I can move it back out. It's a copy/paste from a script where it made sense to have it inside the class...

opengever/maintenance/scripts/export_and_close_dossiers.py
njohner: Ok I did not know that indeed, thanks. I adapted accordingly.

opengever/maintenance/scripts/export_and_close_dossiers.py
njohner: Because I misread your comment, I now also log to a file in the output directory

opengever/maintenance/scripts/export_and_close_dossiers.py
njohner: Thanks, I was actually looking for a better way to copy the files.

njohner:
Thanks for the review. I think I integrated all your suggestions.

lukasgraf (approved):
LGTM! 👍

PR #343, opened 2022-11-01
Some small fixes in the dossier export script after first productive run.

phgross:
- Handle slashes in the dossier title
- Temporary disable the workspaceclient registry flag. Because zopemaster is not able to communicate with the teamraum, the dossier resolving would fail, if the workspaceclient feature is enabled.


lukasgraf (approved):
LGTM 👍

PR #345, opened 2022-11-17
Add script to remove all plone contacts and dossier participations.

phgross:
For some clients we want to enable and integrate our KuB solution. Therefore we need to remove the contact data, which was never be used in a really productive manner. 

For CA-4939

### DIFF
--- a/opengever/maintenance/scripts/contact_removal.py
+++ b/opengever/maintenance/scripts/contact_removal.py
@@ -1,0 +1,86 @@
+"""
+This script is used to remove contacts and related dossier participations.
+
+bin/instance run ./scripts/contact_removal.py
+
+optional arguments:
+  -n : dry-run.
+  -s : site root (used if multiple plone sites exists).
+"""
+
+from ftw.upgrade.progresslogger import ProgressLogger
+from opengever.dossier.behaviors.dossier import IDossierMarker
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/contact_removal.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def remove_contact_participations(self):
+        catalog = api.portal.get_tool('portal_catalog')
+        brains = catalog.unrestrictedSearchResults(
+            object_provides=IDossierMarker.__identifier__)
+        for brain in ProgressLogger(
+                'Remove dossier contact participations.', brains):
+            handler = PloneParticipationHandler(brain.getObject())
+            participations = handler.get_participations()
+            for participation in participations:
```
njohner, line 52: In case this loop is inefficient we could do this directly in the annotations to avoid popping elements one by one from the dictionary:
```
IAnnotations(self.context)['participations'] = PersistentDict({key:value for key, value in IAnnotations(self.context)['participations'] if not key.startswith(u'contact:')})
```

njohner (approved):

opengever/maintenance/scripts/contact_removal.py
phgross: My idea was that the official API also notfiy events, but will adjust if the script is to slow.

PR #351, opened 2022-12-15
Add script to migrate plone contacts to KuB

njohner:
This script is used to migrate the plone contacts to KuB. It is based on the [SQL contact migration script](https://github.com/4teamwork/opengever.maintenance/blob/master/opengever/maintenance/scripts/sql_contact_migration.py). It includes the following steps:
- Create a bundle to be imported into KuB.
- Migrate the dossier participations to KuB participations
- Delete the Plone contacts
- Delete the Contactfolder

Note that the some of the data from the Plone contacts has to be transformed to be imported into KuB, e.g. the country has to be mapped to an ISO Country code, house numbers have to be extracted from the `street` field.
Because certain fields are mandatory on the KuB side (Country and town for an address), we set missing countries to `CH` and skip addresses with missing Town. These are clearly logged and written to a file.

We also do not migrate 3 fields from the plone contacts: `company`, `department` and `function`. We could have created an organization and membership from these fields, but decided against it for now. We do write the fields to the bundle though, so the data would still be around if necessary.

For https://4teamwork.atlassian.net/browse/CA-2605

### DIFF
--- a/opengever/maintenance/scripts/plone_contact_migration.py
+++ b/opengever/maintenance/scripts/plone_contact_migration.py
@@ -28,6 +28,9 @@ from opengever.ogds.base.actor import ContactActor
 from plone import api
 from uuid import uuid4
 from zope.annotation.interfaces import IAnnotations
+from zope.globalrequest import getRequest
+from zope.interface import alsoProvides
+from zope.interface import noLongerProvides
 import argparse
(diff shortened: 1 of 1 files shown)


lukasgraf (approved):
LGTM! 👍

PR #362, opened 2023-03-22
Improve script to fix reference number mappings on repository folders.

njohner:
When a repofolder gets moved (how that happens is still unclear), then we are left with a valid intid in the mapping, which will lead to the reference prefix [still being considered as used](https://github.com/4teamwork/opengever.core/blob/master/opengever/base/adapters.py#L192-L198). Here we update the script cleaning up the mappings to cover that case.

As we currently also have a bug that reference numbers of repofolders do not get indexed properly when the prefix of the parent is changed (https://4teamwork.atlassian.net/browse/CA-4553), and as issues leading to the problems fixed by this script might also have led to incorrect indexing, I also add an option to reindex all repository folders. This makes a rather big transaction which might be slow to commit.

For https://4teamwork.atlassian.net/browse/CA-5497

### DIFF
--- a/opengever/maintenance/scripts/fix_refnum_mappings.py
+++ b/opengever/maintenance/scripts/fix_refnum_mappings.py
@@ -99,14 +99,39 @@ def fix_refnum_mappings(plone):
     print 'Updated a total of {} mappings'.format(n_updated)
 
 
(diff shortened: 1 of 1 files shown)


phgross (approved):
👍

PR #366, opened 2023-04-17
Add scripts to export a subtree as an OGGBundle and transform it for CA-5415

lukasgraf:
Add script to export a subtree as an OGGBundle.

Also includes a script to then transform such a bundle in a way where dossiers get mapped to an already existing repository folder on the target deployment where the bundle will be imported.

Note: The import of a bundle generated this way showed that for `ftw.mail.mail` objects, the filename was not updated based on the title, but left as the filename from the bundle (`file_1.msg`, ...). Therefore those objects need to be fixed in the debug shell afterwards by calling `mail.update_filename()`.

For [CA-5415](https://4teamwork.atlassian.net/browse/CA-5415)

---

These are the steps that were performed for the test export and -import:

1) Export bundle

```bash
bin/instance run src/opengever.maintenance/opengever/maintenance/scripts/export_subtree_bundle.py --without-local-roles --dossiers-with-parent-reference '/areg/ordnungssystem/support-und-ressourcen/informatik-und-telefonie-departement'
```

2) Map dossiers
```bash
bin/instance run src/opengever.maintenance/opengever/maintenance/scripts/map_areg_bdgs_migration_bundle.py areg-export-2023-14-04_19_39_26.oggbundle
```

3) Import bundle
```bash
bin/instance import var/bundles/areg-export-2023-14-04_19_39_26.oggbundle
```

4) Fix mail filenames in debug shell

```python
from DateTime import DateTime
from opengever.maintenance import dm
from plone import api
import transaction

dm()

path = '/areg/ordnungssystem/support-und-ressourcen/informatik-und-telefonie-departement'

catalog = api.portal.get_tool('portal_catalog')
brains = catalog(path=path, portal_type='ftw.mail.mail', created={'query': DateTime('2023/04/19 19:21:11 GMT+2'), 'range': 'min'})

for brain in brains: brain.getObject().update_filename()

transaction.commit()
```

[CA-5415]: https://4teamwork.atlassian.net/browse/CA-5415?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/maintenance/scripts/export_subtree_bundle.py
+++ b/opengever/maintenance/scripts/export_subtree_bundle.py
@@ -1,0 +1,595 @@
+"""Export a subtree as an OGGBundle.
+
+Usage:
+
+export_subtree_bundle.py
+  (--with-local-roles | --without-local-roles)
+  (--dossiers-with-parent-reference | --dossiers-with-parent-guid)
+  <path>
+"""
+
+from Acquisition import aq_inner
+from Acquisition import aq_parent
+from collections import defaultdict
+from collections import OrderedDict
+from datetime import datetime
+from jsonschema import FormatChecker
+from jsonschema import validate
+from opengever.base.interfaces import IOpengeverBaseLayer
+from opengever.base.interfaces import IReferenceNumber
+from opengever.base.schemadump.config import SHORTNAMES_BY_ROLE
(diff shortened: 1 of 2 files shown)


njohner (approved):
LGTM

PR #368, opened 2023-04-24
Bundle export: Skip trashed objects

lukasgraf:
Skip trashed objects, since we don't support setting the trashed status for bundle items during import.

For [CA-5415](https://4teamwork.atlassian.net/browse/CA-5415)


njohner (approved):
LGTM

PR #371, opened 2023-06-14
Add new releases 2023.9 and 2023.10.

njohner:
For https://4teamwork.atlassian.net/browse/CA-5658


njohner:
Pushed directly to master using the new script in opengever-deployments: https://github.com/4teamwork/opengever-deployments/pull/230

PR #372, opened 2023-07-11
Improve script to import propertysheets

njohner:
With this PR we alle the use of references in the json file defining the propertysheets that are imported using the `opengever/maintenance/scripts/add_propertysheets.py` script. We also make a few other improvements in the script.

For https://4teamwork.atlassian.net/browse/CA-5768


phgross (approved):
👍 LGTM

PR #375, opened 2023-09-28
Speedup group migration

njohner:
With this PR we massively speed up the group migration by avoiding the use of `reindexObjectSecurity`. Instead we directly modify the security indexes in plone and in solr, as a group migration is basically a search and replace in those indexes.

For the HBA migration update of the index took around 40 seconds, whereas we never found out how long `reindexObjectSecurity` would take as it never finished, but from discussions with @lukasgraf it seems that even when reindexing just one branch of the repository (the largest branch), reindexing the object security did not finish in a whole weekend.

For https://4teamwork.atlassian.net/browse/CA-5829


lukasgraf (approved):
LGTM! 👍 🎉

PR #376, opened 2023-11-02
Group migration should also work on teamraum deployments.

njohner:
Because the Gever groups are also used in Teamraum, we need to also migrate the groups in a linked teamraum deployment whenever we migrate the groups in a Gever deployment.

For https://4teamwork.atlassian.net/browse/CA-6223


lukasgraf (approved):
LGTM 👍

PR #377, opened 2023-11-02
Fix group migration script and add script to rectify the allowedRolesAndUsers index on already migrated deployments

njohner:
The group migration script contained a bug: when migrating an `old_group` to a `new_group`, if there was already an entry in the forward index for the new_group, we overwrote that entry instead of updating it. This led to the forward entry not containing all the `rids` for such groups.

For https://4teamwork.atlassian.net/browse/CA-6361


lukasgraf (approved):
LGTM 👍

PR #379, opened 2023-12-27
Add active_user_stats script

elioschmutz:
This PR adds a new script to get the active user stats of a deployment.

Run it with

```
bin/instance0 run src/opengever.maintenance/opengever/maintenance/scripts/active_user_stats.py -v
```

to get the following output for a gever installation:

```
{
  "timestamp": "2023-12-27T12:06:59.985791+01:00",
  "total_active_unique_teamraum_users": 0,
  "total_active_unique_gever_users": 267,
  "total_active_unique_spv_users": 291
}
```

or for a teamraum installation:
```
{
  "timestamp": "2023-12-27T12:06:59.985791+01:00",
  "total_active_unique_teamraum_users": 170,
  "total_active_unique_gever_users": 0,
  "total_active_unique_spv_users": 0
}
```

The script uses the OGDS for the calculations. It should be run only once per OGDS.

### DIFF
--- a/opengever/maintenance/scripts/active_user_stats.py
+++ b/opengever/maintenance/scripts/active_user_stats.py
@@ -1,0 +1,93 @@
+"""
+Script that calculates the current active users in total over all admin units an writes it into
+a logfile.
+
+bin/instance0 run src/opengever.maintenance/opengever/maintenance/scripts/active_user_stats.py
+"""
+from datetime import datetime
+from ftw.contentstats.logger import get_log_dir_path
+from logging import FileHandler
+from opengever.maintenance import dm
+from opengever.maintenance.debughelpers import setup_option_parser
+from opengever.meeting import is_meeting_feature_enabled
+from opengever.meeting.model import Committee
+from opengever.ogds.models.group import Group
+from opengever.ogds.models.group import groups_users
+from opengever.ogds.models.org_unit import OrgUnit
+from opengever.ogds.models.user import User
+from opengever.workspace import is_workspace_feature_enabled
+from os.path import join
+from tzlocal import get_localzone
+import json
+import logging
+import pytz
+
+logger = logging.getLogger('opengever.maintenance')
+LOG_TZ = get_localzone()
+
+
+class UserStatsCalculator(object):
+    """Returns statistics for the active users on the system
+    """
+    def get_stats(self):
+        stats = {
+            'total_active_unique_gever_users': 0,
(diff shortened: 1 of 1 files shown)


elioschmutz:
@buchi could you take a look at this script? I don't know how you need the data.

The current script will write the output to a file in `var/log/user-stats-json.log`. I copied the logging from https://github.com/4teamwork/ftw.contentstats/blob/master/ftw/contentstats/logger.py . So it should be the same as the content stats log-file.

Is that enough? Or should I extend `ftw.contentstats` to be able to reuse the code there?

opengever/maintenance/scripts/active_user_stats.py
```diff
@@ -0,0 +1,101 @@
...
+        group_ids = [
+            committee.group_id for committee in
+            Committee.query.filter_by(workflow_state='active')]
+
+        query = self.get_active_users_by_groups_query(group_ids)
+        return query.count()
+
+    def calc_total_active_unique_teamraum_users(self):
+        role_manager = api.portal.get_tool('acl_users').portal_role_manager
+        group_ids = [principal[0] for principal in role_manager.listAssignedPrincipals('WorkspacesUser')]
```
buchi, line 72: Why are you getting the group ids from the role assignments here and not using the same approach as for GEVER users where you get the group ids from the org unit's `users_group_id`?

opengever/maintenance/scripts/active_user_stats.py
```diff
@@ -0,0 +1,101 @@
...
+    handler = FileHandler(path)
+    logger.setLevel(logging.INFO)
+    logger.addHandler(handler)
+    logger.propagate = False
+
+    ts = datetime.utcnow().replace(tzinfo=pytz.utc)
+    stats['timestamp'] = ts.astimezone(LOG_TZ).isoformat()
+
+    value = json.dumps(stats, sort_keys=True)
+    import pdb; pdb.set_trace()
```
buchi, line 100: ✂️

opengever/maintenance/scripts/active_user_stats.py
elioschmutz: You're right. I thought that that's the way for teamraum.

I updated the script 👍

elioschmutz:
@buchi updated 👍 could you take a final look?

buchi (approved):
👍🏻 LGTM

PR #387, opened 2024-03-25
Improve metadata update handling in repository_migration script.

elioschmutz:
We improve the performance and allow to have no blobs when indexing by patching the Catalog and only updating relevant metadata instead the full list of metadata.

The Catalog will always update all available metadata if reindexing an object with `update_metadata` enabled. There is no way to update only specific metadata values of an object. Updating metadata means, that all metadata will be recalcuated for an objects. This process takes a lot of time and is mostly unnecessary. In addition, it requires to have access to the blobs of documents because some of the metadata properties are relying on blob information (filesize). To speed up the migration and to be able to run the migration without the existence of blobs, we'll patch the catalog to only update the relevant metadata items.

### DIFF
--- a/opengever/maintenance/scripts/repository_migration.py
+++ b/opengever/maintenance/scripts/repository_migration.py
@@ -119,6 +119,76 @@ MAPPED_FIELDS["archival_value"].update({
     u'sa': u'archival worthy with sampling',
                                        })
 
+EXPECTED_CATALOG_METADATA_FIELDS = set([
+    'bumblebee_checksum',
+    'changed',
+    'checked_out',
+    'cmf_uid',
+    'contactid',
+    'created',
+    'Creator',
+    'css_icon_class',
+    'delivery_date',
+    'Description',
+    'document_author',
+    'document_date',
+    'email',
+    'email2',
+    'end',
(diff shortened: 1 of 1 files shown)


opengever/maintenance/scripts/repository_migration.py
```diff
@@ -219,6 +284,55 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    """
+
+    def __call__(self):
+        from Products.ZCatalog.Catalog import Catalog
+        from Products.ZCatalog.Catalog import safe_callable
+        from Products.ZCatalog.Catalog import MV
+
+        def recordify(self, obj):
+            # --- Patch ---
+            assert set(self.names) == EXPECTED_CATALOG_METADATA_FIELDS, \
```
buchi, line 307: I would do this only once, not for each recordify call.
Maybe before patching?

buchi (approved):

opengever/maintenance/scripts/repository_migration.py
elioschmutz: Oh yes... sure! Thanks!

PR #391, opened 2024-04-10
Add clusters for walenstadt, pfaefers, quarten & sevelen

blu14x:
Adding configurations for
- walenstadt
- pfaefers
- quarten
- sevelen

to register their interGEVER webactions


lukasgraf (approved):
LGTM 👍
(assuming those URLs are correct :P)

